PR #33, opened 2024-06-18
Integrated translation

### LINKED ISSUE #11: (SP: ) Investigate and connect I18n translations
Should include the translations text for the project
Replace es to uk translation

heshvab:
- Created translation provider with corresponding translation props (messages, locale). The provider itself uses provider provided by our translation library `react-intl`. As you can see, I added it inside of shared folder as it is a globally used thing that at the same time higly related to package. But it can be changed, I am not sure about it.
- Connected translations with application thought provider.
- Also, connected translation with test application, so tests won't fall. For now I just added it inside of our one test, but in future would be good to create some wrapper for tests that wraps children component by providers.
- Replaced spanish example translations by ukrainian.
- Added translations using `FormattedMessage` instead of hardcoded messages (let it be as an example).
- Added translations for dashboard locally, inside of dashboard folder. I think it requires futher discussion, I guess global translations would be easier to use.
- Now you can change language by replacing `en` with `uk` inside of provider props. In future, when we will be implementing language switching we will need to add language context to hold selected language.

The main difference between this library and library we used in SpaceToStudy is that this library uses `FormattedMessage` component for translations, instead of `t` returned by `useTranslation` in `i18-next`. So whenewer you want to add text use `FormattedMessage` with id that equals message `id` in translation constants. There are also a hook `useIntl` that can be used the same way as useTranslation, but as for me, approach with component looks better, also, this approach is recommended by developers of the library. More about this library you can find here https://formatjs.io/docs/react-intl.


D1nviz (approved):

mhevyk (approved):

dmtrth25 (approved):

dmtrth25:
Don't forget select the project
image

OliaHrytsak (approved):

PR #34, opened 2024-06-19
Setup RTK + query

### LINKED ISSUE #26: (SP: 1) Setup RTK + query
Setup store, hooks, slice for products and setup rtk query 
Define custom hooks with typescript

D1nviz:
Created a feature store. Also, I decided to create a folder for asynchronous data `api`,  so it will be the place, where we should put our endpoints.

- Provided some examples of usage.
- Created `createUrlPath` utility function

image

### DIFF
--- a/code/src/constants/requests.ts
+++ b/code/src/constants/requests.ts
@@ -1,0 +1,8 @@
+export const URLS = {
+  todos: {
+    get: "/todos",
+    post: "/todos",
+    put: "/todos",
+    delete: "/todos"
+  } // example of endopoints
+} as const;
--- a/code/src/hooks/use-redux.ts
+++ b/code/src/hooks/use-redux.ts
@@ -1,0 +1,6 @@
+import { useDispatch, useSelector } from "react-redux";
+
+import type { RootState, AppDispatch } from "@/store/store";
+
+export const useAppDispatch = useDispatch.withTypes<AppDispatch>();
+export const useAppSelector = useSelector.withTypes<RootState>();
--- a/code/src/main.tsx
+++ b/code/src/main.tsx
@@ -1,19 +1,25 @@
 import { createRoot } from "react-dom/client";
-import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
+import { Provider } from "react-redux";
 import I18nProivider from "@/shared/i18n/I18nProvider";
+import { StyledEngineProvider, ThemeProvider } from "@mui/material";
+
 import Router from "@/router";
-import '@/styles/scss/global.scss'
-import { theme } from "@/styles/muiTheme"
+import { theme } from "@/styles/muiTheme";
+import { store } from "@/store/store";
+
+import "@/styles/scss/global.scss";
 
 const App = () => {
   return (
-    <StyledEngineProvider injectFirst>
-      <ThemeProvider theme={theme}>
-        <I18nProivider>
-          <Router />
-        </I18nProivider>
-      </ThemeProvider>
-    </StyledEngineProvider>
+    <Provider store={store}>
+      <StyledEngineProvider injectFirst>
+        <ThemeProvider theme={theme}>
+          <I18nProivider>
+            <Router />
+          </I18nProivider>
+        </ThemeProvider>
+      </StyledEngineProvider>
+    </Provider>
   );
 };
 
--- a/code/src/utils/createUrlPath.ts
+++ b/code/src/utils/createUrlPath.ts
@@ -1,0 +1,24 @@
+const createQueryParamsString = (query: Record<string, string>) => {
+  const queryParams = new URLSearchParams();
+
+  Object.entries(query).forEach(([key, value]) => {
+    if (value) {
+      queryParams.append(key, value);
+    }
+  });
+
+  return queryParams.toString();
+};
+
+export const createUrlPath = (URL: string, params = "", query = {}) => {
+  let trimmedUrl = URL;
+  while (trimmedUrl.endsWith("/")) {
+    trimmedUrl = trimmedUrl.slice(0, -1);
+  }
+
+  const queryParams = createQueryParamsString(query);
+  const queryParamsString = queryParams ? `?${queryParams}` : "";
+  const paramsString = params ? `/${params.replace(/^\/+/g, "")}` : "";
+
+  return `${trimmedUrl}${paramsString}${queryParamsString}`;
+};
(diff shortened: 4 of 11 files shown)


code/src/store/features/api/appApi.ts
```diff
@@ -0,0 +1,10 @@
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+export const appApi = createApi({
+  baseQuery: fetchBaseQuery({
+    baseUrl: "http://localhost:5000/",
```
vlad-khrychov: Maybe we need to use `env` to configure `url`-s ? Now we have a dev deployment, does it work there?

D1nviz: Yes, I think so. So I already created an additional issue with setting this file

code/src/hooks/use-redux.ts
```diff
@@ -0,0 +1,6 @@
+import { useDispatch, useSelector, TypedUseSelectorHook } from "react-redux";
+
+import type { RootState, AppDispatch } from "@/store/store";
+
+export const useAppDispatch = () => useDispatch<AppDispatch>();
```
mhevyk: [Official docs](https://redux-toolkit.js.org/usage/usage-with-typescript) provide example with `withTypes` method. What do you think about it?

image

code/src/main.tsx
```diff
@@ -1,19 +1,25 @@
 import { createRoot } from "react-dom/client";
-import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
+import { Provider } from "react-redux";
 import I18nProivider from "@/shared/i18n/I18nProvider";
+import { StyledEngineProvider, ThemeProvider } from "@mui/material/styles";
```
mhevyk: Please double check if we need to import these components from `@mui/material/styles`. It seems like approach from previous mui versions, maybe we can use just `@mui/material`

code/src/store/features/api/appApi.ts
```diff
@@ -0,0 +1,10 @@
...
+    credentials: "include"
+  }),
+  reducerPath: "appApi",
+  endpoints: () => ({})
```
mhevyk: Maybe create test endpoint for our convinience? For example with jsonplaceholder

code/src/store/features/api/appApi.ts
```diff
@@ -0,0 +1,10 @@
+import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+
+export const appApi = createApi({
+  baseQuery: fetchBaseQuery({
+    baseUrl: "http://localhost:5000/",
+    credentials: "include"
+  }),
+  reducerPath: "appApi",
```
mhevyk: Maybe we can use global constants folder for storing this keys instead of hardcoding them?

code/src/hooks/use-redux.ts
D1nviz: I agree, fixed

heshvab (approved):

code/src/main.tsx
D1nviz: fixed

code/src/store/features/api/appApi.ts
D1nviz: used an example url

code/src/store/features/api/appApi.ts
D1nviz: good idea, fixed

code/src/constants/requests.ts
```diff
@@ -0,0 +1,8 @@
+export const URLS = {
+  todos: {
+    get: "/todos",
+    post: "/todos",
+    put: "/todos",
+    delete: "/todos"
+  } // example of endopoints
+};
```
mhevyk: Add `as const` here

code/src/utils/createUrlPath.ts
```diff
@@ -0,0 +1,28 @@
+const createQueryParamsString = (query: { [key: string]: string }) => {
```
mhevyk: Use `Record` here

code/src/constants/requests.ts
D1nviz: fixed

code/src/utils/createUrlPath.ts
D1nviz: Thank you. Fixed

mhevyk (approved):

dmtrth25:
You want to pass `api` folder inside `features` ? Maybe it should be separate folder ?
What do you think ?

D1nviz:
> You want to pass `api` folder inside `features` ? Maybe it should be separate folder ? What do you think ?

I think it's also related to features, so I decided to put it in a features. We decided to use this approach on meeting

code/src/constants/index.ts
dmtrth25, line 1: why it should be `index` file in the main `constants` folder ?

dmtrth25:
> > You want to pass `api` folder inside `features` ? Maybe it should be separate folder ? What do you think ?
> 
> I think it's also related to features, so I decided to put it in a features. We decided to use this approach on meeting

Should to discuss it
But in my mind we should to create separate folder like:
- api
- features -> where inside we'll create folder with some slices etc.

If everything will be inside `features` folder separately

dmtrth25 (approved):

code/src/utils/createUrlPath.ts
```diff
@@ -0,0 +1,28 @@
+const createQueryParamsString = (query: Record<string, string>): string => {
```
mhevyk: Return type here is redundant

code/src/utils/createUrlPath.ts
```diff
@@ -0,0 +1,28 @@
...
+      queryParams.append(key, value);
+    }
+  });
+
+  return queryParams.toString();
+};
+
+export const createUrlPath = (
+  URL: string,
+  params: string | null = "",
```
mhevyk: Do we really need `null` here?

code/src/utils/createUrlPath.ts
D1nviz: Yeah. you're right. Fixed

code/src/utils/createUrlPath.ts
D1nviz: removed

code/src/constants/index.ts
D1nviz: fixed

PR #32, opened 2024-06-18
Define code conventions for our project 

mhevyk:

PR #36, opened 2024-06-19
Feature/16/setup colors and typography

### LINKED ISSUE #16: (SP: ) Setup scss colors and typography

mhevyk:
- Added variables inside `_variables.scss`
- Added scss functions inside `_functions.scss` to handling joining strings and to handle spacing
- Added mixins inside `_mixins.scss` to reuse repeated logic inside our scss
- Added font setup inside `_typography.scss`. Original font from our mockup is not free, so i used `Lato` instead
- Added base styles to body to `global.scss`, `_typography.scss` and `_root.scss`
- Added import alias to `webpack.config.js` to include all scss variables for your usage
- Added `AppTypography` component. It should be used to declare any typography in our project (including headings, captions etc)
- Added `AppButton` component as example of styling usage inside our future components

The more detailed explanation and code conventions will be discussed with our team and added to [docs](https://github.com/idx-academy/spa-orders/pull/32)

The general idea of our styling system is no create custom components (building blocks), that starts of prefix `App`, like `AppButton`, `AppTypography` etc. Then we use mui component inside as a base and add scss styles to it to customize it for our needs. Then we need to use `App`-prefixed component in our project and do not import it from mui directly. For building more complex components, like containers, layouts and pages we should use our wrappers around mui component. In that way we can easily migrate from mui to another ui package without huge refactorings.

**NOTE**: we should discuss with our team if we need to remove `ThemeProvider` from mui, because it will no have any effect on our styling system

### DIFF
--- a/code/src/components/app-typography/AppTypography.tsx
+++ b/code/src/components/app-typography/AppTypography.tsx
@@ -1,0 +1,69 @@
+import { ComponentProps, ReactNode } from "react";
+import { FormattedMessage } from "react-intl";
+import Typography, { TypographyProps } from "@mui/material/Typography";
+import "@/components/app-typography/AppTypography.scss";
+
+type HeadingVariant = "h1" | "h2" | "h3";
+type TextVariant =
+  | "body"
+  | "concept"
+  | "subtitle1"
+  | "subtitle2"
+  | "caption"
+  | "caption-small";
+type AppTypographyVariant = HeadingVariant | TextVariant;
+
+type AppTypographyProps = Omit<TypographyProps, "variant" | "children"> & {
+  variant?: AppTypographyVariant;
+} & (
+    | {
+        translationKey: string;
+        translationProps?: Omit<ComponentProps<typeof FormattedMessage>, "id">;
+        children?: never;
+      }
+    | {
+        translationKey?: never;
+        translationProps?: never;
+        children: ReactNode;
+      }
+  );
+
+function getDefaultComponentTag(variant: AppTypographyVariant) {
+  if (variant.startsWith("h")) {
+    return variant as HeadingVariant;
+  }
+
+  if (variant.startsWith("caption")) {
+    return "span";
+  }
+
+  return "p";
+}
+
+const AppTypography = ({
+  variant = "body",
+  className,
+  component,
+  children,
+  translationKey,
+  translationProps,
+  ...props
+}: AppTypographyProps) => {
+  const TypographyContent = translationKey ? (
+    <FormattedMessage id={translationKey} {...translationProps} />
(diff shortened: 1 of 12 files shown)


code/src/components/app-typography/AppTypography.tsx
```diff
@@ -0,0 +1,52 @@
+import { FormattedMessage } from "react-intl";
+import "./AppTypography.scss";
+import Typography, { TypographyProps } from "@mui/material/Typography";
+import { ReactNode } from "react";
```
D1nviz: organize imports and use alias for importing scss

code/src/components/app-typography/AppTypography.tsx
```diff
@@ -0,0 +1,52 @@
...
+  translationKey,
+  ...props
+}: AppTypographyProps) {
+  return (
+    <Typography
+      component={component ?? getDefaultComponentTag(variant)}
+      className={`spa-typography spa-typography__${variant} ${className}`}
+      {...props}
+    >
+      {translationKey ? <FormattedMessage id={translationKey} /> : children}
```
D1nviz, line 49: move it before return statement

D1nviz: move it before return statement

code/src/components/app-typography/AppTypography.tsx
mhevyk: Fixed

code/src/components/app-typography/AppTypography.tsx
mhevyk: Moved as you mentioned

D1nviz (approved):

OliaHrytsak (approved):

code/src/components/app-typography/AppTypography.tsx
```diff
@@ -0,0 +1,64 @@
...
+  | "subtitle2"
+  | "caption"
+  | "caption-small";
+type AppTypographyVariant = HeadingVariant | TextVariant;
+
+type AppTypographyProps = Omit<TypographyProps, "variant" | "children"> & {
+  variant?: AppTypographyVariant;
+} & (
+    | {
+        translationKey: string;
```
heorhiiss, line 20: There are other props for `<FormattedMessage />` that can be useful, especially `values` which gives us a possibility to dynamically insert values in a message. All of them you can find here: https://formatjs.io/docs/react-intl/components/#formattedmessage

mhevyk: Fixed

heshvab (approved):

dmtrth25 (approved):

PR #40, opened 2024-06-19
Add tests for AppTypography component

### LINKED ISSUE #39: (SP: ) Add tests for AppTypography component

mhevyk:
image

### DIFF
--- a/code/src/components/app-typography/AppTypography.test.tsx
+++ b/code/src/components/app-typography/AppTypography.test.tsx
@@ -1,0 +1,47 @@
+import { render, screen } from "@testing-library/react";
+import AppTypography, {
+  AppTypographyVariant
+} from "@/components/app-typography/AppTypography";
+import { FormattedMessage } from "react-intl";
+
+jest.mock("react-intl", () => ({
+  FormattedMessage: jest.fn()
+}));
+
+const renderAndCheckForTag = (
+  expectedTagName: string,
+  variant?: AppTypographyVariant
+) => {
+  const content = "test content";
+  render(<AppTypography variant={variant}>{content}</AppTypography>);
+
+  const typography = screen.getByText(content);
+  expect(typography).toBeInTheDocument();
+  expect(typography.tagName).toBe(expectedTagName);
+};
+
+describe("AppTypography", () => {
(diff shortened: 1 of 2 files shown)


D1nviz (approved):

OliaHrytsak (approved):

code/src/components/app-typography/AppTypography.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  test("should render translated text if translationKey is provided", () => {
+    const translationKey = "translation.key";
+
+    (FormattedMessage as jest.Mock).mockImplementation(() => translationKey);
+    render(<AppTypography translationKey={translationKey} />);
+
+    const typography = screen.getByText(translationKey);
+    expect(typography).toBeInTheDocument();
+
+    (FormattedMessage as jest.Mock).mockReset();
```
nrevutska: this won't be executed if expect in 43 line fails.
Use mockImplementationOnce in line 39, 
or you can use beforeEach, 
or configure resetting mocks before each test in jest.config

mhevyk: Thank for useful comment. I used `mockImplementationOnce`. I did not want to use `beforeEach`, because it seems redundant as I use mock only in one test. Also it looks like a good idea to use global configuration of clearing mocks for example in `setupTests` file, but in my opition we need a separate issue for it

heshvab (approved):

code/src/components/app-typography/AppTypography.test.tsx
```diff
@@ -0,0 +1,45 @@
+import { render, screen } from "@testing-library/react";
+import AppTypography, {
+  AppTypographyVariant
+} from "@/components/app-typography/AppTypography";
+import { FormattedMessage } from "react-intl";
```
dmtrth25: don't forget about our import convention
- libs
- components,containers...
- styles

should to move above and follow convention

mhevyk: Thanks, fixed

dmtrth25 (approved):

PR #46, opened 2024-06-20
Created footer

### LINKED ISSUE #20: (SP: 1) Create a footer
Should to create a footer at the bottom page

image

heshvab:
- Added Footer component, styles and tests for it
- Small changes inside of `AppTypography` to make possible pass Link inside of component, because previously prop types from passed component were not inherited (so when I did like this `<AppTypography component={Link} to="/about"/>`, there was an error that `to` property does not exist on `AppTypography`)

image

### DIFF
--- a/code/src/components/app-typography/AppTypography.tsx
+++ b/code/src/components/app-typography/AppTypography.tsx
@@ -1,4 +1,4 @@
-import { ComponentProps, ReactNode } from "react";
+import { ComponentProps, ElementType, ReactNode } from "react";
 import { FormattedMessage } from "react-intl";
 import Typography, { TypographyProps } from "@mui/material/Typography";
 import "@/components/app-typography/AppTypography.scss";
@@ -13,13 +13,14 @@ type TextVariant =
   | "caption-small";
 type AppTypographyVariant = HeadingVariant | TextVariant;
 
-type BaseComponent = React.ElementType<any, keyof React.JSX.IntrinsicElements>;
+type BaseComponent = ElementType<any, keyof JSX.IntrinsicElements>;
 
 type AppTypographyProps<T extends BaseComponent = "span"> = Omit<
   TypographyProps<T>,
   "variant" | "children"
 > & {
   variant?: AppTypographyVariant;
+  component?: T;
 } & (
     | {
         translationKey: string;
@@ -53,7 +54,7 @@ const AppTypography = <T extends BaseComponent = "span">({
   translationKey,
   translationProps,
   ...props
-}: { component?: T } & AppTypographyProps<T>) => {
+}: AppTypographyProps<T>) => {
   const TypographyContent = translationKey ? (
     <FormattedMessage id={translationKey} {...translationProps} />
   ) : (
--- a/code/src/components/app-typography/AppTypography.tsx
+++ b/code/src/components/app-typography/AppTypography.tsx
@@ -13,7 +13,12 @@ type TextVariant =
   | "caption-small";
 type AppTypographyVariant = HeadingVariant | TextVariant;
 
-type AppTypographyProps = Omit<TypographyProps, "variant" | "children"> & {
+type BaseComponent = React.ElementType<any, keyof React.JSX.IntrinsicElements>;
+
+type AppTypographyProps<T extends BaseComponent = "span"> = Omit<
+  TypographyProps<T>,
+  "variant" | "children"
+> & {
   variant?: AppTypographyVariant;
 } & (
     | {
@@ -40,7 +45,7 @@ function getDefaultComponentTag(variant: AppTypographyVariant) {
   return "p";
 }
 
-const AppTypography = ({
+const AppTypography = <T extends BaseComponent = "span">({
   variant = "body",
   className,
   component,
@@ -48,7 +53,7 @@ const AppTypography = ({
   translationKey,
   translationProps,
   ...props
-}: AppTypographyProps) => {
+}: { component?: T } & AppTypographyProps<T>) => {
   const TypographyContent = translationKey ? (
     <FormattedMessage id={translationKey} {...translationProps} />
   ) : (
--- a/code/src/layouts/app-footer/AppFooter.constants.tsx
+++ b/code/src/layouts/app-footer/AppFooter.constants.tsx
@@ -1,0 +1,67 @@
+import { Instagram, Twitter, YouTube } from "@mui/icons-material";
+
+export const locationFooterItems = [
+  {
+    label: "footer.location.street"
+  },
+  {
+    label: "footer.location.city"
+  },
+  {
+    label: "footer.location.email",
+    href: "mailto:info@mysite.com"
+  },
+  {
+    label: "footer.location.phone",
+    href: "tel:1234567890"
+  }
+];
+
+export const socialsFooterItems = [
+  {
+    href: "https://facebook.com",
+    element: <YouTube />
+  },
+  {
+    href: "https://instagram.com",
+    element: <Instagram />
+  },
+  {
+    href: "https://twitter.com",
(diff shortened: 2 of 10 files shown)


code/src/components/app-typography/AppTypography.tsx
```diff
@@ -13,7 +13,12 @@ type TextVariant =
   | "caption-small";
 type AppTypographyVariant = HeadingVariant | TextVariant;
 
-type AppTypographyProps = Omit<TypographyProps, "variant" | "children"> & {
+type BaseComponent = React.ElementType<any, keyof React.JSX.IntrinsicElements>;
```
mhevyk: Please use just `ElementType`, do not use `React.`. If you want to use `React.`, just import it from `react` package

code/src/components/app-typography/AppTypography.tsx
```diff
@@ -40,15 +45,15 @@ function getDefaultComponentTag(variant: AppTypographyVariant) {
...
+const AppTypography = <T extends BaseComponent = "span">({
   variant = "body",
   className,
   component,
   children,
   translationKey,
   translationProps,
   ...props
-}: AppTypographyProps) => {
+}: { component?: T } & AppTypographyProps<T>) => {
```
mhevyk: Avoid union types here, use union above

code/src/layouts/app-footer/AppFooter.constants.tsx
```diff
@@ -0,0 +1,67 @@
+import { Instagram, Twitter, YouTube } from "@mui/icons-material";
```
mhevyk: Use default imports

code/src/layouts/app-footer/AppFooter.scss
```diff
@@ -0,0 +1,64 @@
+@import "@/styles/scss/foundation/all";
```
mhevyk: Use `@import "design-system"` instead

code/src/layouts/app-footer/AppFooter.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.footer {
+  padding: spacing(18) 0;
+  background-color: $white;
```
mhevyk: Use combination of `scss` variables and `css` variables. Check for example `AppButton.scss`

code/src/layouts/app-footer/AppFooter.scss
```diff
@@ -0,0 +1,64 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: spacing(2);
+  }
+
+  &__list-title {
+    font-size: $fs-lg;
+    text-decoration: none;
+    margin-bottom: spacing(10);
+    font-weight: 600;
```
mhevyk: Font weight should be used from design system. Maybe consider using `AppTypography`

code/src/layouts/app-footer/AppFooter.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {
+  customerSupportFooterItems,
+  locationFooterItems,
+  policyFooterItems,
+  socialsFooterItems
+} from "@/layouts/app-footer/AppFooter.constants";
+
+import "@/layouts/app-footer/AppFooter.scss";
+
+const AppFooter: FC = () => {
```
mhevyk: We discussed that we won't use `FC`, just type props directly

code/src/layouts/app-footer/AppFooter.tsx
```diff
@@ -0,0 +1,99 @@
...
+        translationKey={item.label}
+        component={item.href ? Link : undefined}
+        to={item.href as string}
+      />
+    </Box>
+  ));
+
+  const socialsItems = socialsFooterItems.map((item) => (
+    <Box component="li" className="footer__socials-item" key={item.href}>
+      <a href={item.href}>{item.element}</a>
```
mhevyk: Use `Link` here

code/src/layouts/app-footer/AppFooter.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  &__list {
+    padding: 0;
+    display: flex;
+    flex-direction: column;
+    gap: spacing(2);
+  }
+
+  &__list-title {
+    font-size: $fs-lg;
```
D1nviz: I would recommend you to use css var.
For example: (also you can look into `AppButton` component)
`var(spa-list-font-small-size, $fs-lg)` or smth like it

code/src/layouts/app-footer/AppFooter.scss
```diff
@@ -0,0 +1,64 @@
...
+    list-style: none;
+  }
+
+  &__list-item-link {
+    text-decoration: none;
+    color: $black;
+    transition: 0.25s color;
+
+    &:hover {
+      color: $dark-alpha;
```
D1nviz: same for this color and hover

code/src/layouts/app-footer/AppFooter.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </Box>
```
D1nviz: Think about using `AppButton` or `AppTypography` with prop `component={Link}`. Maybe it would be better solution

code/src/components/app-typography/AppTypography.tsx
heorhiiss: Thank, fixed

code/src/components/app-typography/AppTypography.tsx
heorhiiss: Thanks, moved it AppTypographyProps

code/src/layouts/app-footer/AppFooter.constants.tsx
heorhiiss: Thanks for noticing, fixed

code/src/layouts/app-footer/AppFooter.scss
heorhiiss: Fixed

code/src/layouts/app-footer/AppFooter.scss
heorhiiss: Thanks, I made it here and in other places the same way as in `AppButton.scss`.

code/src/layouts/app-footer/AppFooter.scss
heorhiiss: Fixed using `AppTypography`

code/src/layouts/app-footer/AppFooter.tsx
heorhiiss: Thanks for noticing, I forgot to delete it. Fixed

code/src/layouts/app-footer/AppFooter.tsx
heorhiiss: Fixed

code/src/layouts/app-footer/AppFooter.scss
heorhiiss: Thanks for suggestion, I fixed using similar to `AppButton` approach here and in other places

code/src/layouts/app-footer/AppFooter.scss
heorhiiss: Fixed

code/src/layouts/app-footer/AppFooter.tsx
heorhiiss: Thanks for suggestion, I used `Link` there. I think simple `Link` would be enough because those elements are just icons without text or some specific styling (there are no background or borders), so no need in `AppButton` or `AppTypography`.

mhevyk (approved):

OliaHrytsak (approved):

D1nviz (approved):

PR #52, opened 2024-06-21
Added .env.local file and dotenv-webpack package to extract envs

### LINKED ISSUE #35: Create and setup .env file
We need to create a file for saving privacy data like keys, api base url,  etc.

heshvab:
- Added `.env.local` to store our environment variables
- Added `code/.env.local` to `.gitignore`
- Added `dotenv-webpack` package to dev dependecies to load envs. Webpack does not support envs by default unlike `vite` (only throught cli when the app is starting), so some custom logic or package is needed. `dotenv-webpack` package is small (19.8 kb), has more than a million weekly downloads and in general recommended to use
- To access environment variables, use `process.env.VAR_NAME`. I set env for RTK base path as an example.
- I used the same approach as on space2study when we have one private file with envs (some projects use two files where one is private and another one is not)


dmtrth25 (approved):

dmtrth25:
Don't forget to handle status

image

D1nviz (approved):

code/package.json
```diff
@@ -38,13 +38,15 @@
...
     "@testing-library/user-event": "^14.5.2",
+    "@types/dotenv-webpack": "^7.0.7",
     "@types/jest": "^29.5.12",
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
     "babel-jest": "^29.7.0",
     "babel-loader": "^9.1.3",
     "babel-plugin-module-resolver": "^5.0.2",
     "css-loader": "^7.1.2",
+    "dotenv-webpack": "^8.1.0",
```
nrevutska, line 49: this should be dev dependency.
How you will access environment variables in the cloud (in prod environment)? Most probably you won't have env.local there.

heorhiiss: Thanks for suggestion. `dotenv-webpack` and `@types/dotenv-webpack` are already in dev dependencies, or am I missing something?

.gitignore
```diff
@@ -16,3 +16,5 @@ code/reports/mutation
 # Mock backend
 mock-backend/node_modules
 
+# Environment variables
+code/.env.local
```
mhevyk: better add all `.env` files to `.gitignore`

heorhiiss: Thanks for suggestion. I used `.env*` to add all `.env` files to gitignore.

mhevyk (approved):

PR #56, opened 2024-06-21
Changed translations structure

heshvab:
- Created folders with translations for `AppBanner` and `AppFooter` (only these components have specific translations). You can notice that I named it `messages`, because it is more descriptive than `i18n` and only messages without any additional logic stored there.
- Added folder `common-messages` inside of `code/shared/i18n` which stors files for common messages. There is only `login.lable` for now. Location of common messages inside of our project's structure can be futher discussed.

Write your suggestions if you have any.

Structure of translations for specific components:
image

Structure of common translations inside of `code/shard/i18n` folder.
image


D1nviz (approved):

code/src/shared/i18n/I18nProvider.tsx
```diff
@@ -8,7 +8,7 @@ type I18nProividerProps = {
 
 const I18nProivider: FC<I18nProividerProps> = ({ children }) => {
   return (
-    <IntlProvider locale="en" defaultLocale="en" messages={messages.en}>
+    <IntlProvider locale="uk" defaultLocale="uk" messages={messages.uk}>
```
mhevyk: Why default locale is changes from `en` to `uk`?

heorhiiss: Very nice observation, thanks. I changed it to test the website with ukrainian language, but forgot to change it back. Fixed.

dmtrth25:
As a conclusion.
We should to create `massage` folder with `en.json` `uk.json` and `index.ts` for exports inside each folder for translations ?
If we have some `common` translations we should to pass it in `shared/i18n/common-messages` ? 
Is everything right ?

OliaHrytsak (approved):

dmtrth25 (approved):

mhevyk (approved):

PR #64, opened 2024-06-21
Added AppIconButton component

### LINKED ISSUE #60: Create AppIconButton component

mhevyk:


D1nviz (approved):

D1nviz (approved):

heshvab (approved):

code/src/components/app-icon-button/AppIconButton.tsx
```diff
@@ -0,0 +1,9 @@
+import IconButton, { IconButtonProps } from "@mui/material/IconButton";
+
+type AppIconButtonProps = IconButtonProps;
+
+const AppIconButton = (props: AppIconButtonProps) => {
```
heorhiiss, line 9: Consider making it generic in a similar to `AppTypography` way, because when I use it like this `<AppIconButton component={Link} to="/" />`, ts throws error because of `to` prop. Icons are often used as links, so it would be useful.

OliaHrytsak (approved):

code/src/components/app-icon-button/AppIconButton.tsx
mhevyk: I created new issue to create `AppLink` component, maybe we can take advantage of composition like following:

```jsx
<AppLink to="/link-to-page">
    <AppIconButton>
         <MyIconComponent />
    </AppIconButton>
</AppLink>
```

What do you think about it?

dmtrth25 (approved):

Issue #63, opened 2024-06-21
Create test-util RenderWithProviders


OliaHrytsak:
To avoid duplication of code in each test file, add RenderWithProviders test-utils that simplifies the rendering of components with all necessary providers. 
This utility ensures that all required contexts, such as Redux, React Router, and any other global providers, are consistently applied across your test suite. Here's how you can implement and use RenderWithProviders

PR #90, opened 2024-06-24
Added route for 'shop all' page

### LINKED ISSUE #83: (SP: ) Create page route for products
When user press 
image 
button should to include route for `all products` like `/products`

Should create after #77

heshvab:
- Added page file
- Added route for it inside of routes
- Added `BrowserRouter` in categories' tests because I added `Link` inside of `Categories` component. Would be good to change it with `renderWithProviders` in #63

### DIFF
--- a/code/src/constants/routes.ts
+++ b/code/src/constants/routes.ts
@@ -1,7 +1,14 @@
 const routes = {
-  home: "/",
-  products: "/products",
-  any: "*"
+  home: {
+    path: "/"
+  },
+  products: {
+    path: "/products",
+    route: "products"
+  },
+  any: {
+    path: "*"
+  }
 } as const;
 
 export default routes;
--- a/code/src/constants/routes.ts
+++ b/code/src/constants/routes.ts
@@ -1,0 +1,7 @@
+const routes = {
+  home: "/",
+  products: "/products",
+  any: "*"
+} as const;
+
+export default routes;
--- a/code/src/constants/routes.ts
+++ b/code/src/constants/routes.ts
@@ -3,8 +3,7 @@ const routes = {
     path: "/"
   },
   products: {
-    path: "/products",
-    route: "products"
+    path: "/products"
   },
   any: {
     path: "*"
--- a/code/src/layouts/app-header/components/categories/Categories.test.tsx
+++ b/code/src/layouts/app-header/components/categories/Categories.test.tsx
@@ -3,7 +3,7 @@ import { BrowserRouter } from "react-router-dom";
 import Categories from "@/layouts/app-header/components/categories/Categories";
 import { categories } from "@/layouts/app-header/components/categories/Categories.constants";
 
-describe("Categories Component", () => {
+describe("Categories component", () => {
   test("render number length of categories", () => {
     render(
       <BrowserRouter>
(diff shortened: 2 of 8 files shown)


code/src/routes/routes.tsx
```diff
@@ -17,6 +18,16 @@ const routes: RouteObject[] = [
...
+  {
+    path: "/products",
+    errorElement: <ErrorPage />,
+    children: [
+      {
+        element: <RootLayout />,
+        children: [{ index: true, element: <ProductsPage /> }]
+      }
+    ]
+  },
```
mhevyk: it is not correct. Our routes should look like following:

```jsx
{
    path: "/",
    errorElement: <ErrorPage />,
    children: [
      {
        element: <RootLayout />,
        children: [
           { index: true, element: <HomePage /> },
           { path: "products", element: <ProductsPage /> }
        ]
      }
    ]
  },
```

OliaHrytsak (approved):

code/src/routes/routes.tsx
D1nviz: I suggest creating object with routes and exporting it as const. For example
```
const routes = {
  home: "/",
  products: "/products",
  notFound ....
 } as const 
```

heorhiiss: Thanks for suggestion. I created routes file in `constants`. It has similar to s2s structure with `path` and `route` properties.

mhevyk (changes requested):

code/src/layouts/app-header/components/categories/Categories.test.tsx
```diff
@@ -1,10 +1,15 @@
 import { render, screen } from "@testing-library/react";
+import { BrowserRouter } from "react-router-dom";
 import Categories from "@/layouts/app-header/components/categories/Categories";
 import { categories } from "@/layouts/app-header/components/categories/Categories.constants";
 
 describe("Categories Component", () => {
```
mhevyk: Why component in uppercase?

code/src/layouts/app-header/components/categories/Categories.tsx
```diff
@@ -1,3 +1,5 @@
+import { Link } from "react-router-dom";
```
mhevyk: Use `AppLink` instead

mhevyk (changes requested):

code/src/layouts/app-header/components/categories/Categories.test.tsx
heorhiiss: It was not written by me. I fixed it.

code/src/layouts/app-header/components/categories/Categories.tsx
heorhiiss: Thanks, used AppLink there.

code/src/routes/routes.tsx
heorhiiss: Thank you, fixed it according to provided example.

mhevyk (approved):

code/src/constants/routes.ts
```diff
@@ -0,0 +1,14 @@
+const routes = {
+  home: {
+    path: "/"
+  },
+  products: {
+    path: "/products",
+    route: "products"
```
D1nviz: Why you are using route property? Do we really need it?

heorhiiss: Well, in this case both properties can be used in the same way, so `route` can be removed. But in some cases, we need absolute path (`path` property) as well as relative path (relative to parent's path, `route` property). `Path` we can use in `Link` for example. `route` property we can use inside of our router, where each route can accept children whose paths must be relative to their parent's path. Maybe this needs better naming, cause just `route` and `path` in this case are not very descriptive.

D1nviz (approved):

dmtrth25 (approved):

PR #98, opened 2024-06-25
Configured webpack for bg images

### LINKED ISSUE #75: Configure webpack to use bg-images inside scss
You need to change the configuration and add the possibility of using background images inside `scss` files

dmtrth25:
image

- now we can use `background-image` inside `scss` not inline at the component also this works for our logic -> changing the image in a few seconds
- replace settings instead `file-loader` and remove this package
- configured `webpack` for that

### DIFF
--- a/code/src/layouts/intro-banner/IntroBanner.tsx
+++ b/code/src/layouts/intro-banner/IntroBanner.tsx
@@ -18,10 +18,7 @@ const IntroBanner = () => {
 
   return (
     <AppBox
-      className="spa-banner-intro"
-      style={{
-        backgroundImage: `url(${introBannerImages[currentImageIndex]})`
-      }}
+      className={`spa-banner-intro spa-banner-intro--image-${currentImageIndex}`}
     >
       <AppContainer maxWidth="lg">
         <AppBox className="spa-banner-intro__wrapper">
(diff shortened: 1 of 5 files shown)


code/src/layouts/intro-banner/IntroBanner.tsx
```diff
@@ -18,10 +18,7 @@ const IntroBanner = () => {
 
   return (
     <AppBox
-      className="spa-banner-intro"
-      style={{
-        backgroundImage: `url(${introBannerImages[currentImageIndex]})`
-      }}
+      className={`spa-banner-intro spa-banner-intro--image-${currentImageIndex}`}
```
D1nviz, line 21: I think `spa-banner-intro__image` would be correct according to BEM methodology. `--` just for modifiers, not elements

dmtrth25: > I think `spa-banner-intro__image` would be correct according to BEM methodology. `--` just for modifiers, not elements

Sure. In my case it's a modifier that indicates a variation of the `spa-banner-intro` block based on the current image index

`__` used for sub elements I guess. It's a main wrapper as well why should I to use sub elements here instead ?

D1nviz: Okay, let it be

D1nviz (approved):

heshvab (approved):

mhevyk (approved):

PR #112, opened 2024-06-26
Create teast-util renderWithProwiders and fix existing imports. Tests for BestSellers

### LINKED ISSUE #63: Create test-util RenderWithProviders

OliaHrytsak:
image
image

Stryker coverage
image

image


code/src/layouts/best-sellers/BestSellers.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { screen } from "@testing-library/react";
+
+import BestSellers from "@/layouts/best-sellers/BestSellers";
+import { useGetProductsQuery } from "@/store/api/productsApi";
+import { renderWithProviders } from "@/utils/test-utils";
+
+import { mockData } from "@/layouts/best-sellers/BestSellers.constants";
+
+jest.mock("../../store/api/productsApi", () => ({
```
D1nviz: can we use alias here?

D1nviz:
Do we need these 2 files? maybe would be better to put them to git ignore?
image

code/config/jest.polyfills.js
```diff
@@ -1,2 +1,3 @@
 // Canvas
+// import "node-fetch";
```
heorhiiss: What does this comment mean?

mhevyk (approved):

code/src/layouts/best-sellers/BestSellers.test.tsx
OliaHrytsak: Yes. Fixed. Thank you.

code/config/jest.polyfills.js
OliaHrytsak: I fogot to remove. Fixed. Thaak you

OliaHrytsak:
> Do we need these 2 files? maybe would be better to put them to git ignore? image

I dont know actually, because we have this files on our develop branch with initial data

heshvab (approved):

dmtrth25 (approved):

D1nviz (approved):

PR #116, opened 2024-06-27
Pushed footer to the bottom of the page

### LINKED ISSUE #92: (SP:) Push footer to bottom
It is considered a good approach to stick footer to the bottom of the page. Please do it.

image

heshvab:
- Pushed footer to the bottom of the page.

image

https://github.com/idx-academy/spa-orders/assets/62070431/fa26d14c-6786-482d-8b05-3192d43b3670


mhevyk (commented):
Excellent work! I just noticed small issue with formatting, in my opinion we should have more commas) If I am not right, please correct me

code/src/layouts/root-layout/RootLayout.scss
```diff
@@ -0,0 +1,9 @@
+.root-layout {
+  display: flex;
+  flex-direction: column;
+  min-height: 100vh;
```
D1nviz: maybe use dvh property? Also, we can add this to our scss variable screen-height and screen-width or smth like that. What do you think?

OliaHrytsak (approved):

heshvab:
> Excellent work! I just noticed small issue with formatting, in my opinion we should have more commas) If I am not right, please correct me

Thanks for noticing, you are right. I experimented with prettier and it used my local settings, instead of project settings. Fixed

code/src/layouts/root-layout/RootLayout.scss
heorhiiss: Thank you. I used `svh` for it because it is static, and with `dvh` there can be unexpected resizes and other bugs. I also added a variable in design system and used it.

D1nviz (approved):

mhevyk (approved):

dmtrth25 (approved):

PR #118, opened 2024-06-27
Created empty modal view

### LINKED ISSUE #105: Create emty modal view and opening logic
image

D1nviz:
https://github.com/idx-academy/spa-orders/assets/118522791/a18cbf4c-0a9b-4ee0-ac14-f199397e559b

### DIFF
--- a/code/src/context/ModalContext.tsx
+++ b/code/src/context/ModalContext.tsx
@@ -1,9 +1,8 @@
 import {
+  PropsWithChildren,
   ReactElement,
-  ReactNode,
   createContext,
   useContext,
-  useRef,
   useState
 } from "react";
 import Dialog from "@mui/material/Dialog";
@@ -11,34 +10,46 @@ import AppBox from "@/components/app-box/AppBox";
 
 type ModalContextType = {
   handleOpenModal: (component: ReactElement) => void;
+  handleToggleModal: (component: ReactElement) => void;
   handleCloseModal: () => void;
 };
 
-type ModalProviderProps = { children: ReactNode };
+type ModalProviderProps = Required<PropsWithChildren>;
 
 const ModalContext = createContext<ModalContextType | null>(null);
 
 const ModalProvider = ({ children }: ModalProviderProps) => {
-  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [modal, setModal] = useState<ReactElement | null>(null);
 
-  const modalRef = useRef<ReactElement | null>(null);
+  const isModalOpen = Boolean(modal);
+
   const handleOpenModal = (component: ReactElement) => {
-    modalRef.current = component;
-    setIsModalOpen(true);
+    setModal(component);
   };
 
   const handleCloseModal = () => {
-    setIsModalOpen(false);
-    modalRef.current = null;
+    setModal(null);
   };
 
+  const handleToggleModal = (component: ReactElement) => {
+    if (modal) {
+      handleCloseModal();
+    } else {
+      handleOpenModal(component);
+    }
+  };
+
+  const modalContent = modal && (
+    <Dialog open={isModalOpen} onClose={handleCloseModal}>
+      <AppBox>{modal}</AppBox>
+    </Dialog>
+  );
+
   return (
-    <ModalContext.Provider value={{ handleOpenModal, handleCloseModal }}>
-      {modalRef.current && (
-        <Dialog open={isModalOpen} onClose={handleCloseModal}>
-          <AppBox>{modalRef.current}</AppBox>
-        </Dialog>
-      )}
+    <ModalContext.Provider
+      value={{ handleOpenModal, handleCloseModal, handleToggleModal }}
+    >
+      {modalContent}
       {children}
     </ModalContext.Provider>
   );
--- a/code/src/context/ModalContext.tsx
+++ b/code/src/context/ModalContext.tsx
@@ -1,0 +1,57 @@
+import {
+  ReactElement,
+  ReactNode,
+  createContext,
+  useContext,
+  useRef,
+  useState
+} from "react";
+import Dialog from "@mui/material/Dialog";
+import AppBox from "@/components/app-box/AppBox";
+
+type ModalContextType = {
+  handleOpenModal: (component: ReactElement) => void;
+  handleCloseModal: () => void;
+};
+
+type ModalProviderProps = { children: ReactNode };
+
+const ModalContext = createContext<ModalContextType | null>(null);
+
+const ModalProvider = ({ children }: ModalProviderProps) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const modalRef = useRef<ReactElement | null>(null);
+  const handleOpenModal = (component: ReactElement) => {
+    modalRef.current = component;
+    setIsModalOpen(true);
+  };
+
+  const handleCloseModal = () => {
+    setIsModalOpen(false);
+    modalRef.current = null;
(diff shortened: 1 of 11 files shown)


dmtrth25:
Don't forget to set status

image

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
+import {
+  ReactElement,
+  ReactNode,
+  createContext,
+  useContext,
+  useRef,
+  useState
+} from "react";
```
dmtrth25, line 8: we try to follow `default` imports is not it ?

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Dialog from "@mui/material/Dialog";
+import AppBox from "@/components/app-box/AppBox";
+
+type ModalContextType = {
+  handleOpenModal: (component: ReactElement) => void;
+  handleCloseModal: () => void;
+};
+
+type ModalProviderProps = { children: ReactNode };
```
dmtrth25: It's looks like `I18nProividerProps` type
Move this `type` to `common` types and reuse it. define only in one place

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const modalRef = useRef<ReactElement | null>(null);
+  const handleOpenModal = (component: ReactElement) => {
+    modalRef.current = component;
+    setIsModalOpen(true);
+  };
+
+  const handleCloseModal = () => {
+    setIsModalOpen(false);
+    modalRef.current = null;
+  };
```
dmtrth25: it can be combined in one `function`. `set` and `modalRef` are repeated
I suggest to create a `fn` with checks condition

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <ModalContext.Provider value={{ handleOpenModal, handleCloseModal }}>
+      {modalRef.current && (
+        <Dialog open={isModalOpen} onClose={handleCloseModal}>
+          <AppBox>{modalRef.current}</AppBox>
+        </Dialog>
+      )}
+      {children}
```
dmtrth25: too long logic here
create variable for `modalRef` here. 
But for now it's not readable

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+type ModalProviderProps = { children: ReactNode };
+
+const ModalContext = createContext<ModalContextType | null>(null);
+
+const ModalProvider = ({ children }: ModalProviderProps) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const modalRef = useRef<ReactElement | null>(null);
+  const handleOpenModal = (component: ReactElement) => {
```
heorhiiss: Usually, we use `handle` prefix with event handlers. In this case the function will be used not as event handler, but just like `handleOpenModal(<SomeComponent/>)`, inside of other functions. Maybe would be better to rename this function and `handleCloseModal` to just `openModal` and `closeModal`? What do you think?

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -9,16 +10,24 @@ import AppLogo from "@/components/app-logo/AppLogo";
 import AppLink from "@/components/app-link/AppLink";
 import AppTypography from "@/components/app-typography/AppTypography";
 
+import { useModalContext } from "@/context/ModalContext";
+
```
dmtrth25, line 14: analyze your unnecessary spaces and remove them

code/src/context/ModalContext.tsx
D1nviz: We decided to use a handle prefix for our handlers. It's a function that handles the opening modal. Also, it has been written in our code conventions 

image

Also provided example from the official MUI documentation
https://mui.com/material-ui/react-dialog/
image

code/src/layouts/modals/auth/AuthModal.tsx
```diff
@@ -0,0 +1,27 @@
+import CloseIcon from "@mui/icons-material/Close";
+
```
dmtrth25, line 2: remove space

code/src/main.tsx
```diff
@@ -2,7 +2,9 @@ import { Provider } from "react-redux";
 import { createRoot } from "react-dom/client";
 import { RouterProvider, createBrowserRouter } from "react-router-dom";
 import { StyledEngineProvider } from "@mui/material";
+
 import I18nProivider from "@/context/I18nProvider";
+import { ModalProvider } from "@/context/ModalContext";
```
dmtrth25, line 7: the same here. Don't forget to follow imports structure and also spaces
The space is only for `styles` in file

code/src/context/ModalContext.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+type ModalProviderProps = { children: ReactNode };
+
+const ModalContext = createContext<ModalContextType | null>(null);
+
+const ModalProvider = ({ children }: ModalProviderProps) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const modalRef = useRef<ReactElement | null>(null);
```
mhevyk: We use two parallel states. I think we can use only one and derive `isModalOpen`, for example like following:

```jsx
const [modal, setModal] = useState<ReactElement | null>(null)

const isModalOpen = Boolean(modal);
```

code/src/context/ModalContext.tsx
D1nviz: Added one more function, for toggling modal

code/src/context/ModalContext.tsx
D1nviz: removed ref

code/src/context/ModalContext.tsx
D1nviz: Yeah, it's a great idea. Thank you!

mhevyk (approved):

code/src/context/ModalContext.tsx
D1nviz: used utility type from `React`
image

code/src/context/ModalContext.tsx
heorhiiss: I read our documentation before commenting. As for MUI examples, they pass those functions inside of event handlers inside of that one component - naming like this makes much sense. But in case of `useModalContext` - it is a reusable hook. In this example, `handleOpenAuthModal` name suits perfectly, but `handleOpenModal` makes it kinda confusing as for me: 
```
const handleOpenAuthModal = () => {
    handleOpenModal(<AuthModal />);
  };
```
As far as I know, `handle` prefix is set for functions which handle directly propagated by user actions like click on a button or scrolling. But there are cases when we need to open modal without user interaction, for example, inside of a `useEffect` when component is rendered for the first time. Then it is already not an action propagated by user. Even in cases, when we need to open modal on button click, we will have `handleButtonClick` or something like this, and inside we will programmatically call function to open modal.

Those are my thoughts. I don't want to be душним because of naming :)

heshvab (approved):

code/src/context/ModalContext.tsx
D1nviz: Thank you for your explanation, it makes sense. I renamed the function according to your suggestion )

dmtrth25 (approved):

PR #130, opened 2024-06-28
Created dropdown with sorting logic

### LINKED ISSUE #117: Implement dropdown for sorting items
Should to create `dropdown` to choose sort property with backend logic

image

dmtrth25:
https://github.com/idx-academy/spa-orders/assets/56305508/21a4325d-30b4-4366-a112-68d241e5e18b

- created reusable `AppDropdown` component and also with sub component for `map` logic
- created sorting which works with backend requests
- changed price without `$` marks. As we discussed `backend` side renders only number from their side and we decided handle this mark vai front side
- handle logic with backend with `RTK Query` via `productsApi`
- decided to use `recommended ` as a default value to fetch all items just like placeholder


code/src/components/app-dropdown/components/AppDropdownOptions.tsx
```diff
@@ -0,0 +1,28 @@
...
+  handleOptionClick
+}: AppDropdownOptionsProps) => {
+  return (
+    <AppBox className="app-dropdown__options">
+      {options.map((option) => (
+        <AppBox
+          key={option.value}
+          className="app-dropdown__option"
+          onClick={(event: MouseEvent) =>
+            handleOptionClick(event, option.value, option.label)
```
OliaHrytsak, line 18: Good work! What do you think about moving this logic above return ?

dmtrth25: > What do you think about moving this logic above return ?

You mean `handleOptionClick` ? But then I can't to get `option` value from the `map` method

OliaHrytsak (approved):

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        options={options}
+        onSelect={mockAppDropdownOnSelect}
+        defaultLabel="Select an option"
+        className="test-dropdown"
+      />
+    );
+  });
+
+  test("renders the default label", () => {
+    expect(screen.getByText("Select an option")).toBeInTheDocument();
```
mhevyk: Move `screen.getByText("Select an option")` to separate variable

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { renderWithProviders } from "@/utils/test-utils";
+
+const mockAppDropdownOnSelect = jest.fn();
+
+const options = [
+  { value: "option1", label: "Label 1" },
+  { value: "option2", label: "Label 2" }
+];
+
+describe("AppDropdown block", () => {
```
mhevyk: Do we need to call it block here? You can just remove it

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  beforeEach(() => {
+    renderWithProviders(
+      <AppDropdown
+        options={options}
+        onSelect={mockAppDropdownOnSelect}
+        defaultLabel="Select an option"
+        className="test-dropdown"
```
mhevyk, line 19: `classname` should not be required here

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(screen.getByText("Select an option")).toBeInTheDocument();
+  });
+
+  test("toggles the dropdown when clicked", () => {
+    const dropdown = screen.getByText("Select an option");
+
+    fireEvent.click(dropdown);
+
+    options.forEach((option) => {
+      expect(screen.getByText(option.label)).toBeInTheDocument();
```
mhevyk: Same here for `screen.getByText(option.label)`

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    fireEvent.click(dropdown);
+
+    options.forEach((option) => {
+      expect(screen.queryByText(option.label)).not.toBeInTheDocument();
```
mhevyk: Same here for `screen.queryByText(option.label)`

code/src/components/app-dropdown/AppDropdown.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+  test("selects an option and calls onSelect", () => {
+    const dropdown = screen.getByText("Select an option");
+
+    fireEvent.click(dropdown);
+
+    const optionToSelect = screen.getByText("Label 2");
+
+    fireEvent.click(optionToSelect);
+
+    expect(screen.getByText("Label 2")).toBeInTheDocument();
```
mhevyk: Same here

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const handleToggleDropdownClick = () => {
+    setIsOpen((prevOpenState) => !prevOpenState);
+  };
+
+  const handleOptionClick = (
+    event: MouseEvent,
+    value: string,
+    label: ReactNode
+  ) => {
+    event.stopPropagation();
```
mhevyk, line 47: Why did you use `event.stopPropagation();` here?

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+    handleSelect(value, label);
+  };
+
+  const handleBlur = (event: FocusEvent<HTMLElement>) => {
+    if (!dropdownRef.current?.contains(event.relatedTarget as Node)) {
```
mhevyk, line 52: Can you please explain why do we use `relatedTarget` here?

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const handleOpenDropdown = () => {
+    return (
+      isOpen && (
+        <AppDropdownOptions
+          handleOptionClick={handleOptionClick}
+          options={options}
+        />
+      )
+    );
+  };
```
mhevyk: Don't use function here and call it later. Just create a variable

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <AppBox
+      ref={dropdownRef}
+      className={`app-dropdown ${className}`}
```
mhevyk: Use `cn` utility here

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+      onClick={handleToggleDropdownClick}
+      onBlur={handleBlur}
+      tabIndex={0}
+    >
+      <AppBox component="span" className="app-dropdown__selected">
+        <AppTypography translationKey="sortBy.label" />
+        {selectedLabel}
+        <AppBox component="span">{iconConditional}</AppBox>
```
mhevyk: Use `AppTypography` instead and choose appropriate variant there

code/src/components/app-dropdown/AppDropdown.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { AppDropdownProps } from "@/components/app-dropdown/AppDropdown.types";
+
+import "@/components/app-dropdown/AppDropdown.scss";
+
+const AppDropdown = ({
+  options,
+  onSelect,
+  defaultLabel,
+  className
+}: AppDropdownProps) => {
```
mhevyk: If we create reusable component, we should use all possible props, like `...props`

code/src/components/app-dropdown/components/AppDropdownOptions.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <AppDropdownOptions
+        options={options}
+        handleOptionClick={mockHandleOptionClick}
+      />
+    );
+  });
+
+  test("should render all options", () => {
+    options.forEach((option) => {
+      expect(screen.getByText(option.label)).toBeInTheDocument();
```
mhevyk: Same here

code/src/pages/products/ProductsPage.constants.tsx
```diff
@@ -0,0 +1,19 @@
+import { FormattedMessage } from "react-intl";
+
+export const sortOptions = [
+  {
+    value: "recommended",
+    label: <FormattedMessage id="productsDefault.label" />
```
mhevyk, line 6: Store only id here and render jsx inside `.tsx` file

code/src/pages/products/ProductsPage.tsx
```diff
@@ -1,23 +1,40 @@
...
+import { Product } from "@/types/product.types";
 import { useGetProductsQuery } from "@/store/api/productsApi";
 
 import "@/pages/products/ProductsPage.scss";
 
 const ProductsPage = () => {
-  const { data: products, isLoading } = useGetProductsQuery();
+  const [sortOption, setSortOption] = useState("");
+  const { data: products, isLoading: productsLoading } =
+    useGetProductsQuery(sortOption);
```
mhevyk: Make `sortOption` optional

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > Do we need to call it block here? You can just remove it

Why I should to remove it ? It's general `describe` block for tests

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > `classname` should not be required here

It doesn't matter for tests I guess.
I pass the options which required for the component

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > If we create reusable component, we should use all possible props, like `...props`

I just used `AppDropdownProps ` type for the props

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Why did you use `event.stopPropagation();` here?

Yes. It's disable `stopPropagation` for my elements to handle logic

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Can you please explain why do we use `relatedTarget` here?

It's used to determine if focus has moved outside the dropdown component
Here is the article can read about it:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Don't use function here and call it later. Just create a variable

Can you explain why usage variable better in this case instead function ?

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Use `AppTypography` instead and choose appropriate variant there

Why I should to use `AppTypography` for icons ?

code/src/pages/products/ProductsPage.constants.tsx
dmtrth25: > Store only id here and render jsx inside `.tsx` file

This approach for type `ReactNode` also make sense. not see the reason to change it.
Is It improve something ?

code/src/pages/products/ProductsPage.tsx
dmtrth25: > Make `sortOption` optional

What do you mean ?
I can just pass empty variable here. It doesn't matter

code/src/components/app-dropdown/AppDropdown.test.tsx
mhevyk: Word `Block` is redundant in my opinion

code/src/components/app-dropdown/AppDropdown.test.tsx
mhevyk: Why can't you just remove it then? It reduces amount of code

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Can you explain in more detail?

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Interesting solution. Great work)

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: You create a function, and then call it, it does not make much sense. For example, you can create variable like following:

```jsx
const a = 1;
```

Or like following:

```jsx
const a = (() => {
    return 1;
})()
```

Or like this:

```jsx
const a = (() => {
    return (() => {
        return 1;
    })();
})()
```

All lines produce same output, but first approach is better in terms of readability and it does not add anything to call stack

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Not for icons, but for rendering `span`  instead

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: If you defined reusable component, it should have a lot of possible use cases. If it is not planned for you, move it to layouts please

code/src/pages/products/ProductsPage.constants.tsx
mhevyk: Definitely yes. If we store only translation key, we will avoid duplicating `FormattedMessage` component multiple times and use it only once. Then if we want to change props for `FormattedMessage`, it would be easier to refactor in the future

code/src/pages/products/ProductsPage.tsx
mhevyk: Please don't be the doesn't matter person. Despite the product is mostly for studing purposes, we should try to make it as good as we can. Moreover, it is not hard to tweek a type and our code will become a little better. All best things consists of a bunch of small things we can do ourselves

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > Word `Block` is redundant in my opinion

Is it so important `AppDropdown block` or `AppDropdown` naming for you ?
It's defined with `block` word because it's block of tests

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > Why can't you just remove it then? It reduces amount of code

This component get `className` as a prop. And I covered it with the tests.
Why is it a problem and I should to remove it ? Don't see the reason for it

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Not for icons, but for rendering `span` instead

I no see the problem to render `icons` with `AppBox`.
You can read about `Typography` and what it related to

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > If you defined reusable component, it should have a lot of possible use cases. If it is not planned for you, move it to layouts please

Are you sure to move it to layouts and where this layout should be? Why ? A lof of possible use cases it's about `AppTypography` variants and some others. It's component no more. You suggest to change different color or some border color for it?) Sure it's possible just pass your `className` as a prop. It's reusable component not layout when you can pass your own styles and rewrite exists, pass the default value.. You can check again the props)

code/src/pages/products/ProductsPage.constants.tsx
dmtrth25: > Definitely yes. If we store only translation key, we will avoid duplicating `FormattedMessage` component multiple times and use it only once. Then if we want to change props for `FormattedMessage`, it would be easier to refactor in the future

I'm not understand. What do you suggest to do in this case ?

code/src/pages/products/ProductsPage.tsx
dmtrth25: > Please don't be the doesn't matter person. Despite the product is mostly for studing purposes, we should try to make it as good as we can. Moreover, it is not hard to tweek a type and our code will become a little better. All best things consists of a bunch of small things we can do ourselves

Give some suggesting to review please and not to replace `AppBox` to `AppTypography` and change mock data for tests

And give me answer for suggestion about `sortOptional` optional. I don't understand what do you mean

code/src/components/app-dropdown/AppDropdown.test.tsx
mhevyk: It is obvious that `describe` defines a block of tests. In your case in is butter oil

code/src/components/app-dropdown/AppDropdown.test.tsx
mhevyk: Please remove it if it is redundant

code/src/pages/products/ProductsPage.constants.tsx
mhevyk: Store only translation key in constants and rename it to `.ts` extension

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > Please remove it if it is redundant

I'm not agree with you. I covered all props value with tests as well. Why is it redundant ? Please explain

code/src/pages/products/ProductsPage.constants.tsx
dmtrth25: > Store only translation key in constants and rename it to `.ts` extension

I already define `ReactNode` type for it. I won't to rewrite types everywhere with this approach. Because for now it look nice and don't see a reason to pass separate key value in constants

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Can you explain in more detail?

Sure. It's propagate to the parent element `AppBox` in my case which has `onClick` event handler. It's would inadvertently toggle dropdown causing it will not close immediately after choosing sort variant opening which is not the intended behavior
using `event.stopPropagation()` it avoid propagate and the component behaves as expected


https://github.com/idx-academy/spa-orders/assets/56305508/e9be63a1-333d-422a-a884-6f35caccffa0

dmtrth25: @mhevyk also I have a good article about it) 
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

code/src/components/app-dropdown/AppDropdown.test.tsx
mhevyk: Okay, let's keep it for now. In the future please try to clean up your redundant code after developing to make our code more readable and less polluted

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Thanks for useful explanation. You actually helped me to understand how it really works

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Just remove redundant `AppBox` if it does not infuence your icon styling, maybe I did not express my thought correctly

code/src/components/app-dropdown/AppDropdown.tsx
mhevyk: Yeap, but in my opinion it is better to spread all props if we need them in some point of time if you want to make component trully reusable and customisable. You can check already existing app components as a reference. I think we can create `containers` folder in the future as on s2s project. `layouts` folder will contain header, footer, navbar, sidebar and different wrappers and boxes and `containers` folder is perfect for components which you introduced. For now it is not an urgent task, but it is worth implementing in the future

code/src/pages/products/ProductsPage.constants.tsx
mhevyk: I don't agree with you. Let's rewrite it. If it is hard for you, i can suggest my help

code/src/pages/products/ProductsPage.tsx
mhevyk: We can try to use object for example passed to `useGetProductsQuery`:

```jsx
const { data: products, isLoading: productsLoading } = useGetProductsQuery({
    sortStrategy: "newest",
    language: "en",
    page: 0,
    ...
});

```

I think your hook will be more configurable from backend side and it will be better to use object to apply filters, sorting, language and so on

I took url from backend as a reference
http://localhost:8080/retail/v1/products?lang=en&page=0&size=8&sort=price,asc

heshvab (approved):

code/src/components/app-dropdown/AppDropdown.test.tsx
dmtrth25: > It is obvious that `describe` defines a block of tests. In your case in is butter oil

Thank you! Removed

code/src/components/app-dropdown/AppDropdown.tsx
dmtrth25: > Use `cn` utility here

Fixed

code/src/pages/products/ProductsPage.constants.tsx
dmtrth25: > I don't agree with you. Let's rewrite it. If it is hard for you, i can suggest my help

We can leave it. and ask our experts. It just your opinion don't force it

mhevyk (approved):

PR #131, opened 2024-06-28
Feature/122/order view page

### LINKED ISSUE #122: Create order view page

D1nviz:
- Added Order Page
- Added OrderItem
- AppTypography font-weight option
- Added Tests
- Created  Order Types

https://github.com/idx-academy/spa-orders/assets/118522791/f246e3ce-5710-486d-bbd5-a0d9c21e2c55

After removing constants it will be 100%
image


OliaHrytsak (approved):

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -81,18 +84,28 @@ const HeaderToolbar = () => {
...
-          </AppIconButton>
+          <AppBox className="header__toolbar-action-icons">
+            <AppLink to={routes.orders.path}>
+              <AppIconButton>
+                <ListAltIcon
+                  className="header__toolbar-cart-icon"
+                  fontSize="large"
+                />
+              </AppIconButton>
+            </AppLink>
```
heorhiiss: Nesting `button` tag inside of `anchor` tag is a bad practice, it fails HTML validator test. I see that `AppIconButton` can be used as link itself by adding `to="your path"` and `component={AppLink}`.

code/src/components/product-card/ProductCard.test.tsx
```diff
@@ -25,7 +25,9 @@ describe("ProductCard component", () => {
   });
 
   test("should render product price", () => {
-    const productPrice = screen.getByText(mockProduct.price);
+    const productPrice = screen.getByText(
+      new RegExp(`${mockProduct.price}\\$`)
```
dmtrth25, line 29: why do you decided to use `RegExp` here ?
It can be easier like ```(`${mockProduct.price}$`)```

code/src/layouts/order-item/components/order-item-details/OrderItemDetails.scss
```diff
@@ -0,0 +1,25 @@
...
+  &__delivery {
+    width: 250px;
+    display: flex;
+    flex-direction: column;
+    gap: spacing(3);
+    padding-block: spacing(2);
+  }
+  &__address {
+    &-label {
+      margin-bottom: 3px;
```
dmtrth25: use `spacing(1)` I think `1px` does't matter. just one approach with `spacing` to define value

code/src/components/app-typography/AppTypography.scss
```diff
@@ -4,10 +4,17 @@
...
+  }
+
   &__h1 {
     font-size: var(--spa-h1-font-size, $fs-h1);
-    font-weight: var(--spa-h1-font-weight, $fw-extra-bold);
     text-transform: var(--spa-h1-text-transform, $tt-none);
+    font-weight: var(--spa-h1-font-weight, $fw-extra-bold);
   }
 
   &__h2 {
```
mhevyk, line 20: Are headings bold by default?

code/src/components/product-card/ProductCard.test.tsx
mhevyk: I hope this solution makes os independent from currency, that we use. It is generally better to prefer regular expression in tests to avoid tight connection between translation of text content and element we want to test. Remember, comma or dollar should not break our tests)

code/src/types/order.types.ts
```diff
@@ -0,0 +1,36 @@
+import { Pageable, Sort } from "@/types/common";
+import { orderStatuses } from "@/constants/orderStatuses";
+import { Product } from "@/types/product.types";
+import { PostAddress, Receiver } from "@/types/delivery.types";
+
+export type OrderStatuses = keyof typeof orderStatuses;
```
mhevyk: Use singular form when declaring types. It is recommended by typescript experts. For example compare following:

```jsx
type Orders = {...}

const orders: Orders[] = []
```

with

```jsx
type Order = {...}

const orders: Order[] = []
```

As you can see, the second approach is basically butter oil.

code/src/components/product-card/ProductCard.test.tsx
mhevyk: I hope this solution makes os independent from currency, that we use. It is generally better to prefer regular expression in tests to avoid tight connection between translation of text content and element we want to test. Remember, comma or dollar should not break our tests)

D1nviz: Yeah. Also in the future, I will replace it with the custom utility function, which will convert numbers into currency

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
D1nviz: Yeah, you're right. Added forward ref to `AppLink` component. Now we can use it like `componen`t prop

code/src/layouts/order-item/components/order-item-details/OrderItemDetails.scss
D1nviz: Replaced by spacing

code/src/components/app-typography/AppTypography.scss
D1nviz: Yes, they are

code/src/types/order.types.ts
D1nviz: Thank you for your advice. Fixed

mhevyk (approved):

heshvab (approved):

dmtrth25 (approved):

PR #135, opened 2024-06-29
Added lazy-loading to pages

### LINKED ISSUE #88: (SP: ) Use React.lazy for importing pages

heshvab:
- Changed ts module to `esnest`. Dynamic imports were introduced in `es2020` and setting esnext or `es2020` or just upper than `es2020` version is needed. I think `esnext` is a good choice because it includes newest features.
- Fixed scss `mapjoin` function. Previously it setted delimiter at the very beginning. Because of this, link for our fonts was not valid and request for them failed. It does not throw any error without `lazy`. But with `lazy`, there was `loading CSS chunk failed` error.
- Added `ignoreOrder` to configuration of `MiniCssExtractPlugin`. With `lazy loading`, `MiniCssExtractPlugin` throws an error, when import order does not equal usage order. As far as I investigated, it is done to properly rewrite `css` rules if the same selector is encountered (basically, if there is `a {rule}` and `a {rule}` again, the second will be applied). I don't think we need ordering error because everything works well without it and usually we do not rely on order of css rules. Besides, it would require us to sort imports according to their usage.
- Added two suspense. Suspense inside of `RootLayout` is for pages inside of layout, so when the page is changing, header and footer will not be suspended (still will be shown). Suspense inside of main is for other pages, which does not have layout, like error page. There are no fallbacks for now, maybe we will add them in future.
- Added lazy loading for pages in `router`.
- Added AppSkeleton and PageLoadingFallback, tests for it. I decided to use skeleton because we will have skeletons on product pages, and people with slow internet will see transition between page loading fallback and product items skeletons. In case of circle loader, it look not very well. Also, there are two variants of skeleton for page loading, one for pages with header and footer and the other one is full screen skeleton (with paddings)

Demonstration (skeleton is that grey thing that appears before content and quickly getting replaced by content)

https://github.com/idx-academy/spa-orders/assets/62070431/1bada516-2c6e-43ee-903e-a059ef240465

### DIFF
--- a/code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
+++ b/code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
@@ -5,8 +5,34 @@ describe("Test PageLoadingFallback", () => {
   test("Should render loading fallback", () => {
     render(<PageLoadingFallback />);
 
-    const circleProgressElement = screen.getByTestId("page-loading-skeleton");
+    const fallbackContainerElement = screen.getByTestId(
+      "page-loading-fallback"
+    );
+    const skeletonElement = screen.getByTestId(
+      "page-loading-fallback-skeleton"
+    );
 
-    expect(circleProgressElement).toBeInTheDocument();
+    expect(fallbackContainerElement).toBeInTheDocument();
+    expect(skeletonElement).toBeInTheDocument();
+  });
+
+  test("Should have proper class when fullScreen is not passed", () => {
+    render(<PageLoadingFallback />);
+
+    const circleProgressElement = screen.getByTestId("page-loading-fallback");
+
+    expect(circleProgressElement).not.toHaveClass(
+      "page-loading-fallback__full-screen"
+    );
+  });
+
+  test("Should have proper class when fullScreen is passed", () => {
+    render(<PageLoadingFallback fullScreen />);
+
+    const circleProgressElement = screen.getByTestId("page-loading-fallback");
+
+    expect(circleProgressElement).toHaveClass(
+      "page-loading-fallback__full-screen"
+    );
   });
 });
--- a/code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
+++ b/code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
@@ -1,0 +1,12 @@
+import { render, screen } from "@testing-library/react";
+import PageLoadingFallback from "@/layouts/page-loading-fallback/PageLoadingFallback";
+
+describe("Test PageLoadingFallback", () => {
+  test("Should render loading fallback", () => {
+    render(<PageLoadingFallback />);
+
+    const circleProgressElement = screen.getByTestId("page-loading-skeleton");
+
+    expect(circleProgressElement).toBeInTheDocument();
+  });
+});
--- a/code/src/layouts/root-layout/RootLayout.tsx
+++ b/code/src/layouts/root-layout/RootLayout.tsx
@@ -4,13 +4,16 @@ import Footer from "@/layouts/footer/Footer";
 import AppBox from "@/components/app-box/AppBox";
 
 import "@/layouts/root-layout/RootLayout.scss";
+import { Suspense } from "react";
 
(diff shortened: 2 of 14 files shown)


code/src/layouts/root-layout/RootLayout.tsx
```diff
@@ -4,13 +4,16 @@ import Footer from "@/layouts/footer/Footer";
...
 import "@/layouts/root-layout/RootLayout.scss";
+import { Suspense } from "react";
 
 const RootLayout = () => {
   return (
     <AppBox className="root-layout">
       <Header />
       <AppBox className="root-layout__content">
-        <Outlet />
+        <Suspense>
```
mhevyk: Please consider adding fallback, you can use `AppLoader` for it

code/src/main.tsx
```diff
@@ -17,7 +18,9 @@ root.render(
     <StyledEngineProvider injectFirst>
       <I18nProivider>
         <ModalProvider>
-          <App />
+          <Suspense>
```
mhevyk: Same here

code/src/layouts/root-layout/RootLayout.tsx
heorhiiss: Thanks, I added fallback there

code/src/main.tsx
heorhiiss: Added fallback here too

mhevyk (approved):

code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { render, screen } from "@testing-library/react";
+import PageLoadingFallback from "@/layouts/page-loading-fallback/PageLoadingFallback";
+
+describe("Test PageLoadingFallback", () => {
+  test("Should render loading fallback", () => {
+    render(<PageLoadingFallback />);
```
dmtrth25, line 6: try to use `beforeEach` as I see this logic repeated for each test

code/src/styles/foundation/_variables.scss
```diff
@@ -115,6 +115,7 @@ $spacing-scale-factor: 0.25rem;
 =============================================*/
 
 $screen-height: 100svh;
+$header-height: 112px;
```
dmtrth25: I guess it shouldn't be here in general variables in our design system variables
I think we can use value for `height` localy for `scss` it will not a problem to define `height` there

D1nviz (approved):

code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
heorhiiss: Thanks for suggestion, in one of those tests I pass different prop, which must not be passed in other tests, I guess `beforeEach` won't suit here. I think some render function that can be created, but I don't see much sense in it because there is not much difference between calling, for example `renderLoadingCallback()` and using `render(<PageLoadingFallback />)`

code/src/styles/foundation/_variables.scss
heorhiiss: Thanks, it makes sense. I moved `header-height` variable to the file where it is used. I still think it is good to store that value inside of a `scss` variable because it is used in `calc` function and without name it is very unclear what we store in that variable.

code/src/layouts/page-loading-fallback/PageLoadingFallback.test.tsx
dmtrth25: > Thanks for suggestion, in one of those tests I pass different prop, which must not be passed in other tests, I guess `beforeEach` won't suit here. I think some render function that can be created, but I don't see much sense in it because there is not much difference between calling, for example `renderLoadingCallback()` and using `render(<PageLoadingFallback />)`

I agree. Creation separate fn to render it is not a good idea. 
I thought maybe to mock some props data and saw that you use only `fullScreen` prop and it will be some useless fn which render your props component handle it and call it for each test

dmtrth25 (approved):

PR #142, opened 2024-06-30
Created a formatPrice util

### LINKED ISSUE #129: Create formatPrice utility
Convert number into a currency
image

D1nviz:
Converts number value into the correct currency
image
image

### DIFF
--- a/code/src/utils/formatPrice.ts
+++ b/code/src/utils/formatPrice.ts
@@ -1,5 +1,7 @@
+type Currency = "USD" | "EUR" | "UAH";
+
 type FormatPriceOptions = {
-  currency?: "USD" | "EUR" | "GBP" | "BDT";
+  currency?: Currency;
   notation?: Intl.NumberFormatOptions["notation"];
 };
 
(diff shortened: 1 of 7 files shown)


code/src/utils/formatPrice.ts
```diff
@@ -0,0 +1,18 @@
+const formatPrice = (
+  price: number,
+  options: {
+    currency?: "USD" | "EUR" | "GBP" | "BDT";
```
mhevyk: Move to separate type for better readability

D1nviz: Moved

code/src/utils/formatPrice.ts
```diff
@@ -0,0 +1,16 @@
+type FormatPriceOptions = {
+  currency?: "USD" | "EUR" | "GBP" | "BDT";
```
mhevyk: What about separate type for currency? I like to move union types like variant into separate type

D1nviz: That's what you mean? Refactored

mhevyk (approved):

heshvab (approved):

dmtrth25 (approved):

PR #156, opened 2024-07-01
Created formatDate util

### LINKED ISSUE #137: Create FormatDate utility function
it should format date from backend `2024-06-27T12:35:14.396Z` to correct format `27.06.2024, 15:35:14`

D1nviz:
image
image

### DIFF
--- a/code/src/utils/format-date/formatDate.ts
+++ b/code/src/utils/format-date/formatDate.ts
@@ -15,7 +15,8 @@ const formatDate = (
     hour: "2-digit",
     minute: "2-digit",
     second: "2-digit",
-    hour12: false
+    hour12: false,
+    timeZone: "Europe/Kiev"
   };
 
   return new Intl.DateTimeFormat(locale, {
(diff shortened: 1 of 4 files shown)


code/src/utils/format-date/formatDate.ts
```diff
@@ -0,0 +1,28 @@
+type FormatDateOptions = {
+  locale?: string;
+  options?: Intl.DateTimeFormatOptions;
+};
+
+const formatDate = (
+  dateString: string,
+  { locale = "en-GB", options = {} }: FormatDateOptions = {}
+) => {
+  const date = new Date(dateString);
```
mhevyk: I suggest extending this function so that it can also accept `Date` and not only date string, because in some cases we may need to pass `Date` directly

You can use something like this:

```jsx
const date = dateLike instanceof Date ? new Date(dateLike) : dateLike;
```

D1nviz: Thank you. Fixed

OliaHrytsak (approved):

mhevyk (approved):

PR #157, opened 2024-07-01
Added tests for layouts folder and utils folder

### LINKED ISSUE #115: Increase tests coverage according to stryker report for layouts and utils
image
image

image
image

image
image

image
image

heshvab:
- Added tests to utils and layouts
- Note that some files are not covered for 100% because of unready implementation like below
- Root layout is uncovered because there is already a separate task for that

Example of uncovered lines that needs to be covered after implementation is ready
image

Mutation coverage
image

image

Test coverage

image

image


code/src/layouts/best-sellers/BestSellers.test.tsx
```diff
@@ -16,64 +16,68 @@ jest.mock("@/store/api/productsApi", () => ({
...
+  error: null
+};
+
+type Options = {
+  data: typeof mockData | null;
+  isLoading: boolean;
+  isSuccess: boolean;
+  isError: boolean;
+  error: null;
+};
```
dmtrth25: I guess maybe it should be `common` type. Which will use on other tests related to `RTK`
What do you think ?

code/src/layouts/intro-banner/IntroBanner.test.tsx
```diff
@@ -2,9 +2,20 @@ import { render, screen } from "@testing-library/react";
...
     const headerText = screen.getByText("introBanner.header");
     expect(headerText).toBeInTheDocument();
   });
+
+  test("should render IntroBanner badge with right properties", () => {
+    const badge =
+      screen.getByText("introBanner.badge").parentElement?.parentElement;
+
+    expect(badge).toBeInTheDocument();
+    expect(badge).toHaveClass("spa-badge__no-rounded");
```
dmtrth25, line 19: I think `toBeInTheDocument()` will be enough to cover in this case for render `IntroBanner` with right properties

code/src/layouts/best-sellers/BestSellers.test.tsx
mhevyk: I suggest moving type first and annotating `defaultOptions` with it like following:

```jsx
const defaultOptions: Options = {
  data: mockData,
  isLoading: false,
  isSuccess: true,
  isError: false,
  error: null
};
```

It helped me to replace values while refactoring a bit faster

dmtrth25: > I suggest moving type first and annotating `defaultOptions` with it like following:
> 
> ```js
> const defaultOptions: Options = {
>   data: mockData,
>   isLoading: false,
>   isSuccess: true,
>   isError: false,
>   error: null
> };
> ```
> 
> It helped me to replace values while refactoring a bit faster

It should be separate `common` type and not inside some `defaultOptions`

heorhiiss: Thanks, I moved it to common types and assigned that type to `defaultOptions` object

heorhiiss: Thanks, I moved it to common types, It can be reused in tests later

code/src/layouts/intro-banner/IntroBanner.test.tsx
heorhiiss: Thanks for suggestion, I added that test because of mutations. By default badge is rounded, so we need to pass `isRounded={false}` inside of `IntroBanner` like following:

```
<AppBadge
            variant="danger"
            size="large"
            isRounded={false}
            badgeContent={badgeContent}
          />
```

Mutation tests mutate it and sets true instead of false. So we need to check whether the badge is rounded, I guess the easiest way to do this is to check its class.

code/src/types/common.ts
```diff
@@ -22,3 +22,11 @@ export type APIError = {
...
   detail: string;
 };
+
+export type RTKQueryReturnState<T = null> = {
+  data: T | null;
+  isLoading: boolean;
+  isSuccess: boolean;
+  isError: boolean;
+  error: null;
+};
```
mhevyk: I suggest also adding optional generic for error

heorhiiss: Thanks for suggestion. I added error generic, it can be useful later

mhevyk (approved):

D1nviz (approved):

OliaHrytsak (approved):

dmtrth25 (approved):

PR #160, opened 2024-07-02
Created NotFound page and change view with layout

### LINKED ISSUE #91: (SP: ) Add more local not found page
When page is not found, user is redirected to `NotFoundPage`, but for now header and footer are not kept. You should modify routes config so that we have global not found fallback (keep existing) and add more local for `RootLayout`

https://github.com/idx-academy/spa-orders/assets/109922101/7f91d61f-3bed-4503-9765-b1f300a6fc7f

dmtrth25:
image

- changed view for `NotFound`
- made changes for `routes` include `*` path with `RootLayout`

Resolved #91


code/src/pages/not-found/NotFoundPage.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { screen } from "@testing-library/react";
+import NotFoundPage from "@/pages/not-found/NotFoundPage";
+import { renderWithProviders } from "@/utils/test-utils";
+
+describe("NotFoundPage", () => {
+  test("renders the 404 title correctly", () => {
+    renderWithProviders(<NotFoundPage />);
+
+    const errorCode = screen.getByText("404");
```
mhevyk, line 9: I suggest also test for heading existance on the page, what do you think? In my opinion, not found page should always have `h1` tag

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
+@import "@design-system";
+
+.not-found-page {
+  display: flex;
+  justify-content: center;
+  text-align: center;
+  padding: spacing(8);
+  background-color: var(--spa-not-found-bgc, $light);
```
mhevyk: Why `bgc`? We always used `background-color` to be more descriptive

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
...
+  justify-content: center;
+  text-align: center;
+  padding: spacing(8);
+  background-color: var(--spa-not-found-bgc, $light);
+  margin-bottom: spacing(2);
+
+  &__content {
+    max-width: spacing(200);
+    padding: spacing(12);
+    background-color: var(--spa-not-found-content-bgc, $white);
```
mhevyk: Same in other places below

code/src/pages/not-found/messages/en.json
```diff
@@ -0,0 +1,5 @@
+{
+  "pageNotFound": "Page Not Found",
+  "notFoundDesc": "The page you are looking for might have been removed, had its name changed, or is temporarily unavailable",
```
D1nviz, line 3: Desc its short name of descending. Name it description

code/src/pages/not-found/NotFoundPage.test.tsx
dmtrth25: > I suggest also test for heading existance on the page, what do you think? In my opinion, not found page should always have `h1` tag

What do you mean ? It's a single test to cover the component which render `404` text. It will be enough
In one case you told too many tests and useless to tests even for each props. For now it's too less tests and should write else. Not understand you. It's test related to `NotFound` component which return `404` code

code/src/pages/not-found/messages/en.json
dmtrth25: > Desc its short name of descending. Name it description

Is it important for key naming ? I just used short convention

code/src/pages/not-found/NotFoundPage.test.tsx
mhevyk: It is semantically correct to have at least one `h1` tag per page

code/src/pages/not-found/messages/en.json
D1nviz: In this case, it will be more informative. Just add a little bit more letters and our lives will become easier :)

code/src/pages/not-found/NotFoundPage.test.tsx
dmtrth25: > It is semantically correct to have at least one `h1` tag per page

Agree with you. Look at `variant`
image

code/src/pages/not-found/messages/en.json
dmtrth25: > In this case, it will be more informative. Just add a little bit more letters and our lives will become easier :)

The such case we also used in `s2s` and everything was ok. To use short naming why not

code/src/pages/not-found/NotFoundPage.test.tsx
mhevyk: Yeah, and with my comment I suggested to add small assertion to check if it true for `NotFound` page

dmtrth25: > Yeah, and with my comment I suggested to add small assertion to check if it true for `NotFound` page

Please check the test again. The test check status code with text for the component with `variant` `h1` as you want. Or what do you mean ?

dmtrth25: > > Yeah, and with my comment I suggested to add small assertion to check if it true for `NotFound` page
> 
> Please check the test again. The test check status code with text for the component with `variant` `h1` as you want. Or what do you mean ?

mhevyk: I mean adding following assertion. Also i suggest renaming `errorCode` to `errorCodeElement`:

```jsx
expect(errorCodeElement.tagName).toBe("H1")
```

dmtrth25: > I mean adding following assertion. Also i suggest renaming `errorCode` to `errorCodeElement`:
> 
> ```js
> expect(errorCode.tagName).toBe("H1")
> ```

Why I should to do so ? It's already cover this component with test..
In my case it's a text not an element

code/src/pages/not-found/NotFoundPage.scss
dmtrth25: > Why `bgc`? We always used `background-color` to be more descriptive

Fixed! Thanks!

code/src/pages/not-found/NotFoundPage.test.tsx
mhevyk: I don't force you, it is just my recommendation

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  &__title {
+    color: var(--spa-not-found-title, $dark-alpha);
+    margin-bottom: spacing(3);
+  }
+
+  &__subtitle {
+    color: var(--spa-not-found-content-bgc, $gray);
+    font-size: var(--spa-subtitle-fs, $fs-xl);
```
mhevyk: Same for `font-size`

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
...
+    margin-bottom: spacing(5);
+  }
+
+  &__description {
+    font-size: var(--spa-description-fs, $fs-lg);
```
mhevyk: Same

code/src/pages/not-found/NotFoundPage.test.tsx
dmtrth25: > I don't force you, it is just my recommendation

I would like to ask you. Which advantages it give us ? Testing element instead the text ??

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
...
+    color: var(--spa-not-found-description, $gray);
+    margin-bottom: spacing(8);
+  }
+
+  &__link-text {
+    color: var(--spa-not-link-text, $white);
```
mhevyk: text-color instead, please be consistent with the rest of the project

code/src/pages/not-found/NotFoundPage.scss
```diff
@@ -0,0 +1,43 @@
...
+
+    &:hover {
+      color: var(--spa-not-link, $dark-alpha);
```
mhevyk: Same

code/src/pages/not-found/NotFoundPage.test.tsx
mhevyk: You agreed with me that we should have at least one `h1` tag per page, so what is the problem?

dmtrth25: > You agreed with me that we should have at least one `h1` tag per page, so what is the problem?

I don't see the reason change test `getByText` to get `h1` with `tagName`
Logic to test the same isn't it ?

mhevyk: You don't have to change `getByText`. Adding extra assertion is not a problem

dmtrth25: > You don't have to change `getByText`. Adding extra assertion is not a problem

I won't to add it. because my test will be enough and not to follow your recommendations... why ?
It changes smth ?
In other case I can add more tests for it. But why ??

dmtrth25: > > You don't have to change `getByText`. Adding extra assertion is not a problem
> 
> I won't to add it. because my test will be enough and not to follow your recommendations... why ?
It changes smth ?

mhevyk (approved):

code/src/pages/not-found/messages/en.json
D1nviz: It's not `s2s`, so try to follow best practices

dmtrth25: > It's not `s2s`, so try to follow best practices

Sure. Show me where it's has written in our rules. I'll rewrite it with best practices :D
For now It's your suggestion not more. Do u agree ??

heshvab (approved):

OliaHrytsak (approved):

D1nviz (approved):

PR #169, opened 2024-07-02
Refactored AppButton so it can be used as link

### LINKED ISSUE #158: Refactor AppButton to make it behave as Link
We need to adjust `AppButton` so that it can become a `AppLink`, because nesting a inside button or button inside a in semantically incorrect and it introduces styling problems, for example if we have `AppLink` and `AppButton` as it's child, then the button can also be clicked and trigger navigation outside button boundaries, that is not correct.

I think the best way to implement it is to use `component` prop in `AppButton`. But if you find better solution, feel free to use it

Also we need to refactor all use cases as following:

image

heshvab:
- Added ability for `AppButton` to behave as link by providing `to` property. I think this way is the most comfortable as we don't need to provide `component` property, only `to`. Other props from Link also can be used.
- Also added `forwardRef` because it can be useful later
- Replaced occurrences of `AppButton`s wrapper in `AppLink`s or vise versa


code/src/components/app-button/AppButton.tsx
```diff
@@ -4,35 +4,54 @@ import cn from "@/utils/cn";
 import { AppButtonProps } from "@/components/app-button/AppButton.types";
 
 import "@/components/app-button/AppButton.scss";
+import { Link, NavLink } from "react-router-dom";
+import { ForwardedRef, forwardRef } from "react";
```
mhevyk: Imports from packages should go first

code/src/components/app-button/AppButton.types.ts
```diff
@@ -10,8 +11,15 @@ type ButtonVariant =
...
   size?: ButtonSize;
   isLoading?: boolean;
 };
+
+export type AppButtonProps = BaseProps &
+  (
+    | ({ isNavLink: true } & NavLinkProps)
+    | ({ isNavLink: false } & LinkProps)
+    | { isNavLink?: boolean; to?: To }
+  );
```
mhevyk: Try to reuse `AppLinkProps`:

```jsx
export type AppLinkProps = Omit<MuiLinkProps, "variant" | "component"> &
  BaseProps &
  (
    | ({ isNavLink: true } & NavLinkProps)
    | ({ isNavLink?: false } & ReactRouterDomLinkProps)
  );
```

code/src/components/app-button/AppButton.tsx
heorhiiss: Thanks, fixed

code/src/components/app-button/AppButton.types.ts
heorhiiss: Thanks, I created reusable type
```
type AppLinkWithNavLinkProps =
  | ({ isNavLink: true } & NavLinkProps)
  | ({ isNavLink?: false } & ReactRouterDomLinkProps)
```
I think that is enough because slightly different logic (for example, `to` must be present in `AppLink` but optional in `AppButton`) requires complex types to be created to make it more reusable.

D1nviz (approved):

OliaHrytsak (approved):

mhevyk (approved):

dmtrth25 (approved):

PR #172, opened 2024-07-03
Refactor utils folder structure

### LINKED ISSUE #152: Refactor utils folder
For now we store `utils` in separate files and have tests folder for all tests. We should keep up same modular structure like in `components` or `layouts` folder, where we create folder in cebab case and component with test in the same folder. It should look like following:

image

Current utils folder structure:

image

mhevyk:
image


OliaHrytsak (approved):

code/src/components/app-accordion/app-accorion-container/AppAccordionContainer.tsx
```diff
@@ -1,6 +1,6 @@
 import Accordion from "@mui/material/Accordion";
 import { AppAccordionProps } from "@/components/app-accordion/app-accorion-container/AppAccordionContainer.types";
-import cn from "@/utils/cn";
+import cn from "@/utils/cn/cn";
```
dmtrth25, line 3: like suggestion maybe rename `cn` file to `index` and import just like `import cn from "@/utils/cn"` instead `cn/cn` just for readability
What do you think ?

dmtrth25 (approved):

D1nviz (approved):

PR #205, opened 2024-07-06
Refactored ProtectedLayout into ProtectedRoute and added role protection to it

### LINKED ISSUE #200: Add role prop to ProtectedLayout

mhevyk:
If we want to create protected route, we should wrap component inside ProtectedRoute component as follows:

```jsx
<ProtectedRoute element={<OrdersPage />} allowedRoles={[ROLES.USER, ROLES.ADMIN]} />
```

`allowedRoles` is required field. This decision was made because it is more intuitive to instantly see what roles we accept even if we explicitly provide `ROLES.USER` role

image

image


D1nviz (approved):

heshvab (approved):

code/src/routes/protected-route/ProtectedRoute.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    expect(Navigate).toHaveBeenCalledWith({ to: routePaths.home.path }, {});
+  });
+
+  test("renders protected element if user is authenticated and has required role", () => {
+    renderComponent({
+      isAuthenticated: true,
+      role: ROLES.ADMIN
+    });
+
+    expect(screen.getByText("Protected Content")).toBeInTheDocument();
```
dmtrth25: `expect` with variable

code/src/routes/protected-route/ProtectedRoute.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    renderComponent({ isLoading: true });
+
+    const loadingComponent = screen.getByText("Loading...");
+    expect(loadingComponent).toBeInTheDocument();
+  });
+
+  test("redirects to unauthorized path if user is not authenticated", () => {
+    renderComponent();
+
+    expect(Navigate).toHaveBeenCalledWith({ to: routePaths.home.path }, {});
```
dmtrth25: When the user is not authenticated in your title. Why it should be `home path` redirect If we at the same page? 
And I suggest to use just `toHaveBeenCalled()` to avoid some questions about it

code/src/routes/protected-route/ProtectedRoute.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  test("redirects to unauthorized path if userDetails is null", () => {
+    renderComponent({ isAuthenticated: true });
+
+    expect(Navigate).toHaveBeenCalledWith({ to: routePaths.home.path }, {});
```
dmtrth25: I think this test useless because you covered `unauthorized` logic with previous test

mhevyk: No, it is required because mutation test will fail

code/src/routes/protected-route/ProtectedRoute.test.tsx
mhevyk: Thanks a lot, I created separate variable for it

code/src/routes/protected-route/ProtectedRoute.test.tsx
mhevyk: I used `forbiddenRedirectPath` and `unauthorizedRedirectPath` instead and I hope it would be better to test if we actually redirect to passed path

dmtrth25 (approved):

PR #206, opened 2024-07-06
Fixed all of the console warnings during tests run

### LINKED ISSUE #189: Fix warning related to redux in unit tests
Please fix warning. I believe it is located in `store/api/appApi.ts`. It will make debugging our tests a lot easier because our console won't be overprinted with text

image

heshvab:
- Fixed mocks that caused redux warnings
- Added `roots: ["src"]` to jest configuration. It specifies root folder inside of which jest searches for test files. This helped to fix `jest-haste-map: duplicate manual mock found: fileMock` error. Also, previously while running unit tests with `npm run test`, test files located inside of `stryker-tmp` were also running, this change excludes them (that folder is a copy of our app that is used by stryker for mutations, it is created when mutation test starts and deleted when it ends, but if mutation test was aborted, it stayed there and slowed unit testing).
- Many changed lines is just prettifying done by prettier.


mhevyk (approved):

mhevyk:
It would be a good addition to this pull request if we could configure `stryker` so that it removes `tmp` folder when we abort test running. What do you think about this idea?

heshvab:
> It would be a good addition to this pull request if we could configure `stryker` so that it removes `tmp` folder when we abort test running. What do you think about this idea?

That's a good good idea, I will add it

D1nviz (approved):

OliaHrytsak (approved):

heshvab:
> It would be a good addition to this pull request if we could configure `stryker` so that it removes `tmp` folder when we abort test running. What do you think about this idea?

I couldn't add it for now( Stryker does not have out-of-the-box support for some actions when we terminate mutation tests run.  I tried to use some setup script that listens to termination using `process.on('SIGINT')`, but it throws error that the folder is currently used when I delete it, I couldn't figure out how to solve it yet. For now, let it be this way

mhevyk:
> > It would be a good addition to this pull request if we could configure `stryker` so that it removes `tmp` folder when we abort test running. What do you think about this idea?
> 
> I couldn't add it for now( Stryker does not have out-of-the-box support for some actions when we terminate mutation tests run. I tried to use some setup script that listens to termination using `process.on('SIGINT')`, but it throws error that the folder is currently used when I delete it, I couldn't figure out how to solve it yet. For now, let it be this way

Okay, you already did a great job!

PR #210, opened 2024-07-06
Added ProductsGrid component

### LINKED ISSUE #171: Make reusable component for grid of products
Now products are displayed on Products page and on a home page. Probably they will be displayed in other places too. Would be good to make some reusable container that uses a grid to display cards

heshvab:
- Added component
- Tests for the component
- Used it in those places where it can be used


code/src/components/products-grid/ProductsGrid.tsx
```diff
@@ -0,0 +1,31 @@
...
+  isLoading = false,
+  loadingItemsCount = 5
+}: ProductsGridProps) => {
+  const skeletonCards = repeatComponent(<ProductSkeleton />, loadingItemsCount);
+
+  const productCards = products?.map((product: Product) => (
+    <ProductCard key={product.id} product={product} />
+  ));
+
+  const gridItems = isLoading ? skeletonCards : productCards;
```
mhevyk: Can we also handle error case here? We can render some error fallback instead of empty value

code/src/components/products-grid/ProductsGrid.types.ts
```diff
@@ -0,0 +1,8 @@
+import { Product } from "@/types/product.types";
+
+export type ProductsGridProps = {
+  products: Product[] | undefined;
```
mhevyk: Use optional operator  here instead

heorhiiss: I purposely used it like `Product[] | undefined`. This way we always need to pass something, because there is no point in this component without passing products (just `<ProductsGrid />` will throw an error). `Undefined` is added so we don't need to make checks like `data?.content ?? []` in parent component, instead we can just pass `data?.content`. Usually this component will be used together with rtk query hook and data there is always `undefined` initially.

code/src/components/products-grid/ProductsGrid.tsx
heorhiiss: I guess it can be added

code/src/components/products-grid/ProductsGrid.types.ts
mhevyk: Okay, then please add comment above this line to explain this solution for other teammates and make our code more understandable. Comments are the best tool for such staff

code/src/components/products-grid/ProductsGrid.tsx
heorhiiss: Added ability to pass `isError` property and simple fallback message

mhevyk (approved):

D1nviz:
It would be wonderful if you would add some error fallback for these products. Smth like "Products not found", because now we don't have it
image

heshvab:
> It would be wonderful if you would add some error fallback for these products. Smth like "Products not found", because now we don't have it image

I have already added it after Maksym's comment. The component accepts `isError` boolean and renders error message based on it.
image

D1nviz (approved):

OliaHrytsak (approved):

dmtrth25 (approved):

PR #233, opened 2024-07-08
Added multiple env files

### LINKED ISSUE #191: Add multiple webpack and .env files for different environments (development, production and test)
It already became a pain to switch between mock backend url and real backend url. In the future we will most likely have more variables and more environments. I suggest creating files file following:

image

We need to use external package, one of the most popular solution is `dotenv`, personally I also used `dotenvx`, that has following syntax: `dotenvx run --env-file=.env.production -- node index.js`
We already have `.env.local` file, so we can merge these approaches

I think we need to create multiple webpack configs for different environments

heshvab:
- Added multiple webpack configurations.
- Added development, production and test `.env` files support. 
- Renamed `env.local` to just `.env`. I did it because all of our env files are local so `.local` prefix makes no sense. This file is used for general envs which are the same in different environments.
- Since we already had `dotenv-webpack`, I didn't install another library.
- Env files are loaded from inside of webpack configuration files, so used env depends on configuration file we specify when run our application (specified inside of scripts). On `npm run start` `.env.development` envs are used, `npm run start:prod` uses `.env.production` envs and `npm run start:test` uses `.env.test` envs.
- Deleted commented configuration inside of `webpack.config`

- Write if you have suggestions for enhancements


code/package.json
```diff
@@ -3,8 +3,10 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "start": "webpack serve --config ./config/webpack.config.js --mode development",
-    "build": "webpack --config ./config/webpack.config.js --mode production",
+    "start": "webpack serve --config ./config/webpack.development.config.js --mode development",
+    "start:prod": "webpack serve --config ./config/webpack.production.config.js --mode production",
+    "start:test": "webpack serve --config ./config/webpack.test.config.js --mode development",
+    "build": "webpack --config ./config/webpack.production.config.js --mode production",
```
mhevyk: I'm curious why we use `--mode` flag here. I would be glad to hear more details about it

mhevyk (approved):

mhevyk:
It `webpack-dotenv`'s config there are a lot of useful options. What are your thoughts about them?

image

code/package.json
heorhiiss: It specifies mode argument that is passed to webpack configuration in case we create configuration as function. I guess that argument was not used before and is not used now inside of configuration. I just left it as it was

heshvab:
> It `webpack-dotenv`'s config there are a lot of useful options. What are your thoughts about them?
> 
> image

Good you mentioned it. I guess `systemvars` can be useful (gives possibility to get already mentioned `NODE_ENV` and some other envs) as well as `defaults` for base `.env` file. I guess safe mode would be good for real-world project, but since we will not have many envs, I am not sure about it. As for others, I don't think these are useful for us.

D1nviz (approved):

dmtrth25 (approved):

PR #234, opened 2024-07-09
Dashboard button with tooltip

### LINKED ISSUE #230: Add button for dashboard (icon on header)
image

OliaHrytsak:
-Added dashboard button and custom tooltip component

https://github.com/idx-academy/spa-orders/assets/127191734/e6222740-79d5-465c-b2bd-ef66c264250d

image


code/src/components/app-tooltip/AppTooltip.types.tsx
```diff
@@ -0,0 +1,8 @@
+import { ReactNode } from "react";
+
+import { TooltipProps } from "@mui/material/Tooltip";
+
+export type AppTooltipProps = Omit<TooltipProps, "title"> & {
```
mhevyk: Also omit `placement` and `arrow` props

code/src/store/slices/userSlice.ts
```diff
@@ -103,5 +103,7 @@ export const useIsAuthLoadingSelector = () =>
   useAppSelector((store) => store.user.isLoading);
 export const useUserDetailsSelector = () =>
   useAppSelector((store) => store.user.userDetails);
+export const useIsShopManagerSelector = () =>
+  useAppSelector((store) => store.user.userDetails?.role === "ROLE_SHOP_MANAGER");
```
mhevyk: The better solution will be either creating `useRoleSelector` with role argument or just using existing `useUserDetailsSelector` and destructuring role

OliaHrytsak: Thank you! Changed into useUserDetailsSelector

code/src/components/app-tooltip/AppTooltip.types.tsx
OliaHrytsak: Removed placement as redundant
Omited arrow

D1nviz:
Add a little bit more spacing between the icons, also I think it would be better to make them smaller

OliaHrytsak:
> Add a little bit more spacing between the icons, also I think it would be better to make them smaller

There was redundant AppBox and because of that the spacing was smaller

D1nviz (approved):

dmtrth25 (approved):

code/src/components/app-tooltip/AppTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+import "@/components/app-tooltip/AppTooltip.scss";
+
+const AppTooltip = forwardRef<HTMLDivElement, AppTooltipProps>(
+  function AppTooltip({ titleTranslationKey, children, ...props }, ref) {
+    const translatedTitle = titleTranslationKey ? (
+      <AppTypography
+        className="spa-app-tooltip"
+        translationKey={titleTranslationKey}
+      />
+    ) : undefined;
```
mhevyk: Why do we have nested `AppTooltip` here?

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -32,10 +35,13 @@ const HeaderToolbar = () => {
   const { openModal } = useModalContext();
   const isAuthenticated = useIsAuthSelector();
   const isLoadingAuth = useIsAuthLoadingSelector();
+  const userDetails = useUserDetailsSelector();
   const dispatch = useAppDispatch();
 
   const [searchValue, setSearchValue] = useState("");
 
+  const { role } = userDetails || {};
```
mhevyk: It is not the best solution here, because role can be null and we will be forces write this code over and over again. Makes sense to create useUserRoleSelector, that can look like following:

```jsx
export const useUserRoleSelector = () =>
  useAppSelector((store) => {
    const userDetails = store.user.userDetails;

    if (userDetails === null) {
      return null;
    }

    return userDetails.role;
  });
```

role can be `null` if user is not authorized, so we should take it into account

heshvab (approved):

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
OliaHrytsak: Fixed like you suggests

code/src/components/app-tooltip/AppTooltip.tsx
OliaHrytsak: Fixed

mhevyk (approved):

PR #240, opened 2024-07-09
Add admin dashboard tabs

### LINKED ISSUE #222: Create empty tabs for dashboard
Add empty tabs with switch logic (use search params)

mhevyk:
image

Tests:
image

image


code/src/layouts/dashboard-tabs/DashboardTabs.tsx
```diff
@@ -0,0 +1,67 @@
...
+              tabFromQuery === tab.name && "dashboard-tabs__label-item--active"
+            )}
+          >
+            {tab.icon}
+            <AppTypography
+              component="span"
+              translationKey={tab.labelTranslationKey}
+            />
+          </AppBox>
+        ))}
```
D1nviz: I suggest you move it above the return statement

mhevyk: Moved. I also thought about creating separate `DashboardTabComponent`, I tried to do it, but it involves strange state splitting regarding `searchParams`, because we can use logic in either in parent or child component because search params can be accessed anywhere, it's like context in this case. What are you thoughts about it?

code/src/layouts/dashboard-tabs/DashboardTabs.scss
```diff
@@ -0,0 +1,44 @@
+@import "@design-system";
+
+.dashboard-tabs {
```
OliaHrytsak, line 3: Don't forget about spa prefix)
And what do you think about changing the range of tabs? 
I mean - Products, Orders, Users - according to priority for business and maiby frequency of use

code/src/layouts/dashboard-tabs/DashboardTabs.constants.tsx
```diff
@@ -0,0 +1,46 @@
...
+  name: DashboardTabName;
+  icon: ReactNode;
+  content: ReactNode;
+};
+
+export const DASHBOARD_TAB_NAMES = {
+  USERS: "users",
+  ORDERS: "orders",
+  PRODUCTS: "products"
+} as const;
```
dmtrth25, line 24: add translations for `tabs`

code/src/layouts/dashboard-tabs/DashboardTabs.scss
mhevyk: We use `spa` prefix only for components, prefixed with `App`. About rearranging, it is useful remark, i will do it

code/src/layouts/dashboard-tabs/DashboardTabs.constants.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export type DashboardTabName =
+  (typeof DASHBOARD_TAB_NAMES)[keyof typeof DASHBOARD_TAB_NAMES];
```
dmtrth25: maybe `keyof typeof DASHBOARD_TAB_NAMES` will be enough for this type ?

code/src/layouts/dashboard-tabs/DashboardTabs.scss
```diff
@@ -0,0 +1,44 @@
...
+      background-color: var(
+        --spa-dashboard-tabs-hover-background-color,
+        $ghost-alpha
+      );
+    }
+
+    &--active {
+      border-color: var(--spa-dashboard-tabs-active-color, $purple);
+      background-color: var(
+        --spa-dashboard-tabs-active-bg-color,
```
dmtrth25: `bg` full name as you made above

image

code/src/layouts/dashboard-tabs/DashboardTabs.tsx
```diff
@@ -0,0 +1,67 @@
...
+        tabFromQuery === tab.name && "dashboard-tabs__label-item--active"
+      )}
+    >
+      {tab.icon}
+      <AppTypography
+        component="span"
+        translationKey={tab.labelTranslationKey}
+      />
+    </AppBox>
+  ));
```
dmtrth25: I guess it should be another component for `map` item.
This approach used for `product` and `cart` item as well

dmtrth25:
@mhevyk What do you think to set `background` as on the `product` page and `cart` e.g ?
Just two block with this background for tabs and items. Like `PageWrapper` and inside block with your `class` definition which you already passed to `PageWrapper` ?

code/src/layouts/dashboard-tabs/DashboardTabs.constants.tsx
mhevyk: Translations are already added, these are names, that are used as `key` for `map` and for query params value, for example `/dashboard?tab=users`

code/src/layouts/dashboard-tabs/DashboardTabs.constants.tsx
mhevyk: No, because we need values of object. I think we can try to create reusable type, I will try to do it

code/src/layouts/dashboard-tabs/DashboardTabs.tsx
mhevyk: Please check comment https://github.com/idx-academy/spa-orders/pull/240#discussion_r1671633361 and tell your opinion about it

code/src/layouts/dashboard-tabs/DashboardTabs.scss
mhevyk: Discussed. I reordered it like so: orders - products - users

code/src/layouts/dashboard-tabs/DashboardTabs.constants.tsx
mhevyk: I created type `export type ExtractValues<T> = T[keyof T];` that can be reused across the project. I refactored code snippet like follows:

```jsx
export type DashboardTabName = ExtractValues<typeof DASHBOARD_TAB_NAMES>;
```

code/src/layouts/dashboard-tabs/DashboardTabs.scss
mhevyk: Yeap, fixed

OliaHrytsak (approved):

D1nviz (approved):

mhevyk:
> @mhevyk What do you think to set `background` as on the `product` page and `cart` e.g ? Just two block with this background for tabs and items. Like `PageWrapper` and inside block with your `class` definition which you already passed to `PageWrapper` ?

Let's add separate issue for it. It would be good to implement

heshvab (approved):

dmtrth25 (approved):

PR #243, opened 2024-07-09
Created cart drawer

### LINKED ISSUE #218: Implement Drawer for cart item
When the user click `Add to cart` on the current product. Should be `Drawer` on the right side without change quantity
Make reusable via `modal-context`. Should work with backend data

image

heshvab:
- Added drawer context. Initially I thought to use modal provider for it but I I think separate providers are better in perspective in case we will have dialog that opens from inside of a drawer. Besides, separation makes code less complicated. 
- Added cart drawer. For now with mocked data.
- Moved modal provider and drawer provider to RootLayout. I did it because previously it was located higher in component tree than RouterProvider and react-router's links didn't work inside drawer. 
- Changed Product card a little so drawer opens after clicking Add to cart.
- I am going to create tasks for tests for cart drawer.

https://github.com/idx-academy/spa-orders/assets/62070431/a037a46d-17dd-4878-b584-c170f0118e40


mhevyk (commented):
Ii my point of view animation of showing a drawer is too fast. There is an [API for drawer](https://mui.com/material-ui/api/drawer/), you can configure `transitionDuration`

mhevyk:
And will tests for `DrawerContext` be added within this PR?

heshvab:
> And will tests for `DrawerContext` be added within this PR?

I guess no. I thought to add them after adding items to cart will be implemented, together with some other tests related to cart

code/src/layouts/cart-drawer/CartDrawer.scss
```diff
@@ -0,0 +1,50 @@
+@import "@design-system";
+
+.cart-drawer {
+  width: 400px;
+  display: flex;
+  flex-direction: column;
+  height: 100svh;
```
D1nviz: use the `$screen-height` instead

code/src/layouts/cart-drawer/CartDrawer.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  &__close-button {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    color: var(--spa-black-background-color, $white);
+  }
+
+  &__title {
+    font-size: $fs-lg;
```
D1nviz: I suggest setting `subtitle2` (it uses 20px size)  variant to the `AppTypography` and if you need bold for this text, you can use `fontWeight` prop which includes `extra-bold` variant. For semantic you can use `h2` component

code/src/layouts/cart-drawer/message/index.ts
```diff
@@ -0,0 +1,4 @@
+import en from "./en.json";
+import uk from "./uk.json";
```
D1nviz: use alias for imports

code/src/components/product-card/ProductCard.tsx
```diff
@@ -4,45 +4,45 @@ import AppBox from "@/components/app-box/AppBox";
...
+          component="img"
+          src={product.image}
+        />
+        <AppBox className="spa-product-card__description">
+          <AppTypography>{product.description}</AppTypography>
         </AppBox>
       </AppBox>
-    </AppLink>
+      <AppBox>
+        <AppTypography component={AppLink} to="/" variant="caption">
```
OliaHrytsak: Why are you decided to move link there? Maybe it would be better user experience to wrap into link img+name?

code/src/context/DrawerContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const openDrawer = (component: ReactElement) => {
+    setIsOpen(true);
+    setDrawer(component);
+  };
+
+  const closeDrawer = () => {
+    setIsOpen(false);
+  };
+
+  const toggleDrawer = (component: ReactElement) => {
```
OliaHrytsak, line 35: maybe we can awoid this if/else ?

`drawer ? closeDrawer() : openDrawer(component);`

code/src/components/product-card/ProductCard.tsx
heorhiiss: We have button `Add to cart` inside of a `ProductCard` and if link wraps entire `ProductCard` - by clicking on a button we also click on a link, so user is navigated which is not desired behaviour when user clicks on a `Add to card` button. Probably it can be solved with `stopPropagation`, but that is in general a bad practice to have button inside of a link. So for now I made product title as a link.

code/src/context/DrawerContext.tsx
heorhiiss: Thanks for noticing, fixed

code/src/layouts/cart-drawer/message/index.ts
heorhiiss: Thanks, I fixed it

code/src/layouts/cart-drawer/CartDrawer.scss
heorhiiss: Thanks, I used suggested approach in this case

code/src/layouts/cart-drawer/CartDrawer.scss
heorhiiss: Thanks, I forgot we have such variable despited added it myself) fixed

heshvab:
> Ii my point of view animation of showing a drawer is too fast. There is an [API for drawer](https://mui.com/material-ui/api/drawer/), you can configure `transitionDuration`
I slowed animation speed to 400ms (was 300ms). Could you take a look? I think that is optimal value because it is not too fast and not too slow at the same time.

code/src/layouts/cart-drawer/cart-drawer-item/CartDrawerItem.scss
```diff
@@ -0,0 +1,49 @@
+@import "@design-system";
+
+.cart-item {
```
OliaHrytsak, line 3: Don't forget about spa prefix

heorhiiss: If I remember correctly, we use `spa` prefix only for components, prefixed with `App`?

code/src/components/product-card/ProductCard.tsx
OliaHrytsak: I mean img+title , not all card

heorhiiss: Sorry for misunderstanding, I added link around img and title

mhevyk:
image

1. This button is full height. I think it would be better to make it smaller.
2. Please add `object-fit: cover`-like staff, because images from backend can differ and it will cause ui bugs
3. Add `text-truncate` mixin for name of product, because name can be very long and it would be better to make our layout more stable and independent of content we get from the server

OliaHrytsak (approved):

heshvab:
> image
> 
> 1. This button is full height. I think it would be better to make it smaller.
> 2. Please add `object-fit: cover`-like staff, because images from backend can differ and it will cause ui bugs
> 3. Add `text-truncate` mixin for name of product, because name can be very long and it would be better to make our layout more stable and independent of content we get from the server

Changed button
Added object fit and tested it on different images
Added text truncate and tested it

image

mhevyk (approved):

dmtrth25 (approved):

PR #248, opened 2024-07-10
Added ability to delete and add product to the cart

### LINKED ISSUE #246: Logic to add items to cart saved on a backend
Add logic that gives possibility for authorized users to save products inside of a cart on a backend side.

heshvab:
- Changed backend so it sends mocked id to the fronted. We need it to make requests to the backend.
- Added cartApi slice, including tags.
- Added rtk query requests for deletion and addition
- Added two endpoints on mock-backend 
- Tests and logic for unauthorized users will be added later


dmtrth25 (approved):

code/src/components/products-container/ProductsContainer.tsx
```diff
@@ -21,8 +24,14 @@ const ProductsContainer = ({
   loadingItemsCount = 5,
   errorMessage = "errors.somethingWentWrong"
 }: ProductsContainerProps) => {
+  const user = useUserDetailsSelector();
+
+  const [addToCart] = useAddToCartMutation();
```
mhevyk, line 29: How do we handle loading state?

code/src/components/products-container/ProductsContainer.tsx
```diff
@@ -36,8 +45,18 @@ const ProductsContainer = ({
 
   const skeletonCards = repeatComponent(<ProductSkeleton />, loadingItemsCount);
 
-  const handleAddToCart = () => {
-    openDrawer(<CartDrawer />);
+  const handleAddToCart = async (product: Product) => {
+    try {
+      if (user?.id) {
+        await addToCart({ productId: product.id, userId: user.id }).unwrap();
+        openDrawer(<CartDrawer />);
```
mhevyk, line 52: Why do we use `unwrap` here?

code/src/components/products-container/ProductsContainer.tsx
```diff
@@ -36,8 +45,18 @@ const ProductsContainer = ({
...
+      }
+    } catch (e) {
+      openSnackbar({
+        variant: "error",
+        messageTranslationKey: "cart.itemAddition.fail"
+      });
```
mhevyk: Do we have conflict with global error snackbar?

code/src/layouts/cart-drawer/CartDrawer.tsx
```diff
@@ -9,17 +9,47 @@ import AppIconButton from "@/components/app-icon-button/AppIconButton";
...
 
 import "@/layouts/cart-drawer/CartDrawer.scss";
 
 const CartDrawer = () => {
-  const cartItems = mockCartItems.map((item) => (
-    <CartDrawerItem key={item.productId} {...item} />
-  ));
+  const user = useUserDetailsSelector();
+
+  const [removeItem] = useRemoveFromCartMutation();
```
mhevyk, line 23: Same loading state here

code/src/components/products-container/ProductsContainer.tsx
heorhiiss: This is used when user clicks add to cart button inside of ProductCard (demonstrated below). I think loading indicator is redundant there. In case of successful adding to the cart, drawer with new item opens, so user knows that it was added. Otherwise snackbar is shown.

image

code/src/components/products-container/ProductsContainer.tsx
heorhiiss: Because of `unwrap`, result is returned immediately. In this case I need error to be thrown for `try catch` block to work. Without `unwrap` error is handled internally and `try catch` block would not work.

code/src/layouts/cart-drawer/CartDrawer.tsx
heorhiiss: I am not sure how to display loading indicator for the user in this case. Maybe you can suggest something?

image

code/src/components/products-container/ProductsContainer.tsx
heorhiiss: As far as I see from state changes, global error snackbar handles error and sets according state. Then immediately my snackbar state is setted. Because of transition and instant setting of my snackbar state, message from global snackbar is not being seen

mhevyk: Interseting solution! I think is is caused by `unwrap` in your case

code/src/layouts/cart-drawer/CartDrawer.tsx
mhevyk: Yeah, maybe we can leave it for now and discuss if we need this feature

code/src/components/products-container/ProductsContainer.test.tsx
```diff
@@ -64,6 +64,15 @@ const mockProducts: Product[] = [
   }
 ];
 
+jest.mock("@/hooks/use-snackbar/useSnackbar", () => ({
+  __esModule: true,
+  default: jest.fn(() => ({ openSnackbar: () => {} }))
```
mhevyk, line 69: Why not just `jest.fn()`?

code/src/components/products-container/ProductsContainer.tsx
```diff
@@ -36,8 +45,18 @@ const ProductsContainer = ({
...
 
-  const handleAddToCart = () => {
-    openDrawer(<CartDrawer />);
+  const handleAddToCart = async (product: Product) => {
+    try {
+      if (user?.id) {
+        await addToCart({ productId: product.id, userId: user.id }).unwrap();
+        openDrawer(<CartDrawer />);
+      }
+    } catch (e) {
```
mhevyk: Remove `e`, it is not used

code/src/constants/requests.ts
```diff
@@ -12,5 +14,11 @@ export const URLS = {
   orders: {
     get: "/v1/orders/user",
     post: "/v1/orders"
+  },
+  cart: {
+    post: ({ userId, productId }: CartManagementParams) =>
+      `/v1/users/${userId}/cart/${productId}`,
+    delete: ({ userId, productId }: CartManagementParams) =>
+      `/v1/users/${userId}/cart/items/${productId}`
```
mhevyk, line 22: I suggest using `satistfies` keyword for stricter type limitations, because we want object, which keys are also object with name as a key and string or function that returns a string as a value:

I suggest following approach:

```jsx
type UrlPath = string | ((...tokens: string[]) => string);

export const URLS = {
  auth: {
    signUp: "/auth/sign-up",
    signIn: "/auth/sign-in"
  },
  products: {
    get: "/v1/products",
    post: "/v1/products",
    put: "/v1/products",
    delete: "/v1/products"
  },
  orders: {
    get: "/v1/orders/user",
    post: (id) => /v1/users/${id}/orders
  }
} as const satisfies Record<string, Record<string, UrlPath>>;
```

Maybe you can adjust it to make function arguments more generic, not `string[]` as I demonstrated

code/src/layouts/best-sellers/BestSellers.test.tsx
```diff
@@ -17,6 +17,15 @@ jest.mock("@/store/api/productsApi", () => ({
...
 }));
 
+jest.mock("@/hooks/use-snackbar/useSnackbar", () => ({
+  __esModule: true,
+  default: jest.fn(() => ({ openSnackbar: () => {} }))
+}));
+
+jest.mock("@/store/api/cartApi", () => ({
+  useAddToCartMutation: jest.fn(() => [jest.fn(), {}])
+}));
```
mhevyk, line 27: Same here (`jest.fn()`)

code/src/layouts/cart-drawer/CartDrawer.tsx
```diff
@@ -9,17 +9,47 @@ import AppIconButton from "@/components/app-icon-button/AppIconButton";
...
+
+  const handleRemoveItem = async (product: CartItem) => {
+    try {
+      if (user?.id) {
+        await removeItem({
+          userId: user.id,
+          productId: product.productId
+        }).unwrap();
+      }
+    } catch (e) {
```
mhevyk: Same here with `e`

code/src/components/products-container/ProductsContainer.tsx
```diff
@@ -36,8 +45,18 @@ const ProductsContainer = ({
...
+      openSnackbar({
```
mhevyk: Please consider using `openSnackbarWithTimeout` here if we need autohiding functionality. Same in other places

code/src/pages/products/ProductsPage.test.tsx
```diff
@@ -5,6 +5,15 @@ import { useGetProductsQuery } from "@/store/api/productsApi";
...
 import renderWithProviders from "@/utils/render-with-providers/renderWithProviders";
 
+jest.mock("@/store/api/cartApi", () => ({
+  useAddToCartMutation: jest.fn(() => [jest.fn(), {}])
+}));
+
+jest.mock("@/hooks/use-snackbar/useSnackbar", () => ({
+  __esModule: true,
+  default: jest.fn(() => ({ openSnackbar: () => {} }))
+}));
```
mhevyk, line 15: Same

code/src/store/api/cartApi.ts
```diff
@@ -0,0 +1,26 @@
...
+import { CartManagementParams } from "@/types/cart.types";
+
+const cartApi = appApi.injectEndpoints({
+  endpoints: (build) => ({
+    addToCart: build.mutation<void, CartManagementParams>({
+      query: (args) => ({
+        url: URLS.cart.post(args),
+        method: httpMethods.post
+      }),
+      invalidatesTags: [rtkQueryTags.CART]
```
mhevyk, line 14: Do we need to use tags here? As far as I know, tags are commonly used on `GET` queries. If I am mistaken, please correct me

D1nviz (approved):

code/src/store/api/cartApi.ts
heorhiiss: Tags are used to make automatic refetch (`GET`) on some action (`POST`, `PATCH `or `DELETE`). Now we do not have `GET` request, but soon we will have it. `invalidatesTags` accepts tags, which should be refetched after mutation. So we will provide the same tag to our query for cart and it will be automatically refetched on created in my PR requests.

code/src/components/products-container/ProductsContainer.test.tsx
heorhiiss: Without `() => ({ openSnackbar: () => {} })`. Test falls with `Cannot read properties of undefined (reading 'openSnackbar')`. I guess that is because inside of a component, returned by hook values is destructured.

code/src/layouts/best-sellers/BestSellers.test.tsx
heorhiiss: The same thing, test fall without providing array as a mocked return value.

code/src/store/api/cartApi.ts
mhevyk: Interesting. Then we should also add tags to existing rtk endpoints

code/src/components/products-container/ProductsContainer.tsx
heorhiiss: Thanks, removed

code/src/layouts/cart-drawer/CartDrawer.tsx
heorhiiss: Thanks, removed

code/src/components/products-container/ProductsContainer.tsx
heorhiiss: Thanks, used suggested approach

code/src/pages/products/ProductsPage.test.tsx
heorhiiss: Test falls without it, already explained in other comment

code/src/constants/requests.ts
heorhiiss: Maybe would be good to create an issue to add `satisfies` keyword for our constant data like router paths, URLs, etc. For now I couldn't add it because of unexpected errors I couldn't resolve in a normal way. Code like this throws error for example:
```
type UrlFunctionParam = number | string | Record<string, string | number>;
type UrlFunction = (param: UrlFunctionParam) => string;
type UrlPath = string | UrlFunction;

export const URLS = {
  auth: {
    signUp: "/auth/sign-up",
    signIn: "/auth/sign-in"
  },
  products: {
    get: "/v1/products",
    post: "/v1/products",
    put: "/v1/products",
    delete: "/v1/products"
  },
  orders: {
    get: "/v1/orders/user",
    post: "/v1/orders"
  },
  cart: {
    post: ({ userId, productId }: CartManagementParams) =>
      `/v1/users/${userId}/cart/${productId}`,
    delete: ({ userId, productId }: CartManagementParams) =>
      `/v1/users/${userId}/cart/items/${productId}`
  }
} as const satisfies Record<string, Record<string, UrlPath>>;
```

OliaHrytsak (approved):

mhevyk (approved):

PR #255, opened 2024-07-11
Add containers folder and refactor imports, updated import sort rules, fix package.json scripts

### LINKED ISSUE #153: Add containers folder
For now we have `layouts` folder for complex components. We should create separation between them and create `containers` folder like was on Space2Study project.

In layouts folder we should store following:

- Headers
- Footers
- Navbars
- Sidebars
- Layouts, containers and wrappers (`RootLayout` or `PageWrapper` for example)

All other sections should go into containers folder. It will help us to create good separation for our code

The main idea under layouts is to store repeatable pieces of ui, mostly without any logic and for react-router layouts for example. Containers can use logic inside of them and they are generally less reusable

mhevyk:
- Added `containers` folder
- Added containers folder to `prettier` to sort correctly
- Added scripts to `package.json` to use prettier
- Fix webpack mode configuration
- Update readme
- Formatted all files according to new prettier rules


code/.prettierrc
```diff
@@ -1,20 +1,21 @@
...
-  "arrowParens": "always"
-}
+{
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "importOrder": [
+    "^(.*)react(.*)$",
+    "<THIRD_PARTY_MODULES>",
+    "^@/App$",
+    "^@/containers/(?!.*\\.scss$)(.*)$",
+    "^@/layouts/(?!.*\\.scss$)(.*)$",
```
D1nviz: Move layouts above the containers. It is more global things

D1nviz (approved):

heshvab (approved):

code/.prettierrc
mhevyk: Fixed

PR #258, opened 2024-07-11
Added get orders for dashboard

D1nviz:
Added getAdminOrders logic

- Added a new endpoint for it in the mock backend
- Refactored page wrapper to a more suitable variant
- Added a new UI part for the dashboard and small enhancements.

image
image

PR #259, opened 2024-07-11
Created logic for get products for cart of current user

### LINKED ISSUE #245: Implement logic to get cart items from backend
Should to get cart items with backend endpoints

dmtrth25:
Implemented `GET` request cart items for current user included their `userId`

https://github.com/idx-academy/spa-orders/assets/56305508/632b615b-8b47-41e5-93c6-bd5acbe330c6

- added `PageWrapper` just for view for larger screen sizes
- created `getCartItems` endpoint
- some integration with our mocked `backend`
- correct logic with `Drawer` and changing quantity will be implemented in the future

**Resolved: #245**


code/src/store/api/cartApi.ts
```diff
@@ -2,10 +2,17 @@ import { rtkQueryTags } from "@/constants/api-tags";
...
 import { appApi } from "@/store/api/appApi";
-import { CartManagementParams } from "@/types/cart.types";
+import { CartManagementParams, CartType } from "@/types/cart.types";
 
 const cartApi = appApi.injectEndpoints({
   endpoints: (build) => ({
+    getCartItems: build.query<CartType, number>({
+      query: (userId) => ({
+        url: URLS.cart.get({ userId }),
+        method: httpMethods.get
```
D1nviz: It's a redundant line. Build. query always uses the get method.

code/src/types/common.ts
```diff
@@ -42,3 +42,7 @@ export type PaginationParams = {
 export type RTKQueryMockState<TData = null, TError = null> = Partial<
   RTKQueryReturnState<TData, TError>
 >;
+
+export type UserIdType = {
```
D1nviz: Add this type to our UserTypes. Name it UserId and use it for existing types
image

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,113 @@ type CartItemsType = {
...
+              </AppBox>
             </AppBox>
+            <AppButton
+              className="spa-order-summary__button"
+              variant="contained"
+              size="medium"
+              data-testid="createOrderButton"
+            >
+              <AppTypography translationKey="createOrder.label" />
+            </AppButton>
```
D1nviz: make this button disabled if the order items quantity is 0 or smth like it

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,113 @@ type CartItemsType = {
...
-
-  const cartItemsBlock = cartItems.map((item, index) => (
-    <CartItem
-      key={item.name}
-      item={item}
-      onQuantityChange={handleQuantityChangeWrapper(index)}
-      onDelete={handleOnDeleteWrapper(index)}
-    />
+  const cartItemsBlock = cartItems?.items.map((item: CartItem) => (
+    <CartItem key={item.productId} item={item} />
```
D1nviz: Maybe create some fallback for it if cart items.length is 0. What do you think?

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,113 @@ type CartItemsType = {
...
+              translationKey="myCart.label"
+              data-testid="myCartLabel"
             />
-            <AppBox className="spa-order-summary__row spa-order-summary__total-line">
+            {cartItemsBlock}
+          </AppBox>
+          <AppBox className="spa-cart-page__order-summary">
+            <AppTypography
+              className="spa-cart-page__order-summary--label"
+              variant="h1"
```
OliaHrytsak: Smth strainge with variant and component
If we have variant h2 for what reasone component h2?

code/src/store/api/cartApi.ts
dmtrth25: > It's a redundant line. Build. query always uses the get method.

Removed

code/src/types/common.ts
dmtrth25: > Add this type to our UserTypes. Name it UserId and use it for existing types image

I decided to pass this type to `common` because this type uses for cart items and also orders the same type
So it's `common`

code/src/pages/cart/CartPage.tsx
dmtrth25: > make this button disabled if the order items quantity is 0 or smth like it

When the quantity will be 0 you'll see the text `The cart is empty!`
Also this logic will be relate with changing quantity

code/src/pages/cart/CartPage.tsx
dmtrth25: > Maybe create some fallback for it if cart items.length is 0. What do you think?

I added condition to check length

code/src/pages/cart/CartPage.tsx
dmtrth25: > Smth strainge with variant and component If we have variant h2 for what reasone component h2?

changed to `component="h1"`

code/src/types/common.ts
D1nviz: `UserId` is related exactly to the `User` entity. If you are using it with other entities it does not mean that this type is common. A common type is for general things. Look at the name of this type, the name of this type starts with `User` . Also, you can ask the team if you do not trust me ) According to the naming, we are not using ending  `Type` for union types

code/src/pages/cart/CartPage.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  test("renders loading state", () => {
+    mockUseGetCartItemsQuery.mockReturnValue({
+      data: null,
+      isLoading: true,
+      error: false
+    });
+
+    renderWithProviders(<CartPage />);
```
mhevyk: Make helper function to reuse it, because if is repeated

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,127 @@ type CartItemsType = {
...
+                <AppTypography
+                  component="p"
+                  translationKey="free.label"
+                  data-testid="freeLabel"
+                />
+              </AppBox>
               <AppTypography
-                className="spa-order-summary__total"
-                variant="subtitle2"
                 component="p"
```
mhevyk: Same

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,127 @@ type CartItemsType = {
...
+              data-testid="orderSummaryLabel"
+            />
+            <AppBox className="spa-order-summary__details">
+              <AppBox className="spa-order-summary__row">
+                <AppTypography
+                  component="p"
+                  className="spa-order-summary__text"
+                  translationKey="subtotal.label"
+                  data-testid="subtotalLabel"
+                />
```
mhevyk: `p` is already default component, so you can remove it

code/src/types/common.ts
dmtrth25: > `UserId` is related exactly to the `User` entity. If you are using it with other entities it does not mean that this type is common. A common type is for general things. Look at the name of this type, the name of this type starts with `User` . Also, you can ask the team if you do not trust me ) According to the naming, we are not using ending `Type` for union types

It's `common` type and reusable. Is not it ? 
Otherwise we can ask our mentors and experts

code/src/pages/cart/CartPage.test.tsx
dmtrth25: > Make helper function to reuse it, because if is repeated

I tried but can't. Show me your suggestion for it please and I'll rewrite it

code/src/pages/cart/CartPage.tsx
dmtrth25: > `p` is already default component, so you can remove it

Thank you. Removed

code/src/pages/cart/CartPage.tsx
dmtrth25: > Same

What do you want ? helper-function ?? For what ?

code/src/pages/cart/CartPage.test.tsx
mhevyk:
```jsx

const renderWithMock = ({data = null, isLoading = false, error = false}) => {
    mockUseGetCartItemsQuery.mockReturnValue({
      data,
      isLoading,
      error
    });

    renderWithProviders(<CartPage />);
}
```

code/src/pages/cart/CartPage.tsx
mhevyk: You removed `component`. It is correct

code/src/pages/cart/CartPage.test.tsx
dmtrth25: > ```js
> const renderWithMock = ({data = null, isLoading = false, error = false}) => {
>     mockUseGetCartItemsQuery.mockReturnValue({
>       data,
>       isLoading,
>       error
>     });
> 
>     renderWithProviders(<CartPage />);
> }
> ```

It just function for my case. And why we not created it before..
Create please reusable function. I'm not see the reason used it in one place

mhevyk:
Please disable the remove item from cart button when the amount of items is 1 because it does nothing when we click it

code/src/pages/cart/CartPage.test.tsx
mhevyk: You use similar logic in 3 tests. This should be refactored. Check other tests, we always try to reuse logic in tests

heshvab (approved):

code/src/pages/cart/CartPage.test.tsx
dmtrth25: > You use similar logic in 3 tests. This should be refactored. Check other tests, we always try to reuse logic in tests

Just for your 2 tests it will be ok ? Huh. 3 it's not 2 tests)
Reconsider your review approach and try to follow your own suggestions
FYI if you have similar `renderWithProviders` just pass it into `beforeEach` block and if it's not depend with mock different responses data from hook etc.

I won't to create utility function because that there is no sense in this..

image

mhevyk: You provided an example that confirms your point. I will propose to consider another example. Take a look on `DashboardTabs.test.tsx` for example:

```jsx
const renderWithMockSearchParams = (params?: Record<string, string>) => {
  (useSearchParams as jest.Mock).mockReturnValue([
    new URLSearchParams(params),
    mockSetSearchParams
  ]);

  renderWithProviders(<DashboardTabs />);
};
```

Same with `ProductsPage`, `OrdersPage` written by @D1nviz, `ProductsContainer` and `BestSellers` by @GeorgeShvab, also `useSignIn`, `useSignUp` for example. So, that confirms, that in all places where there was reusable code that differs very little, a setup or render local function was added. Ideally yes, my example could be refactored. But the important note is that one line with `renderWithProviders` in example from my code that you provided does not make a big deal for readability concerns but in your case you just copied and pasted 7 lines of code over and over again (as much as 3 times). You contradict yourself, because you claim that you should avoid code repetition.

In my opinion it should be refactored and I even provided and example on how you should do it.

P.S. I will add rule `you should avoid code duplication as much as possible in your code and your tests` and `try to keep our tests clear and create setup or render local functions to improve code readability` into our readme if it is not obvious for you

mhevyk (changes requested):

code/src/pages/cart/CartPage.test.tsx
dmtrth25: > You provided an example that confirms your point. I will propose to consider another example. Take a look on `DashboardTabs.test.tsx` for example:
> 
> ```js
> const renderWithMockSearchParams = (params?: Record<string, string>) => {
>   (useSearchParams as jest.Mock).mockReturnValue([
>     new URLSearchParams(params),
>     mockSetSearchParams
>   ]);
> 
>   renderWithProviders(<DashboardTabs />);
> };
> ```
> 
> Same with `ProductsPage`, `OrdersPage` written by @D1nviz, `ProductsContainer` and `BestSellers` by @GeorgeShvab, also `useSignIn`, `useSignUp` for example. So, that confirms, that in all places where there was reusable code that differs very little, a setup or render local function was added. Ideally yes, my example could be refactored. But the important note is that one line with `renderWithProviders` in example from my code that you provided does not make a big deal for readability concerns but in your case you just copied and pasted 7 lines of code over and over again (as much as 3 times). You contradict yourself, because you claim that you should avoid code repetition.
> 
> In my opinion it should be refactored and I even provided and example on how you should do it.
> 
> P.S. I will add rule `you should avoid code duplication as much as possible in your code and your tests` and `try to keep our tests clear and create setup or render local functions to improve code readability` into our readme if it is not obvious for you

Changed. But please also follow this rules from your side as I talked about `beforeEach` block in your tests case

mhevyk: Looks much better, thanks! What about mine test? Should I create separate issue for it now or we can fix it in connection to some PR?

code/src/pages/cart/CartPage.tsx
```diff
@@ -19,118 +22,120 @@ type CartItemsType = {
 };
 
 const CartPage = () => {
-  const [cartItems, setCartItems] = useState<CartItemsType[]>(items);
+  const { id: userId } = useUserDetailsSelector() || {};
+  const {
+    data: cartItems,
+    error,
+    isLoading: cartItemsLoading
+  } = useGetCartItemsQuery(userId!);
```
mhevyk, line 29: This can be refactored like I wrote for @D1nviz's PR to avoid `|| {}` and `userId!`. Seems like he is not available for now. Maybe you will merge it faster, so you can also fix it. What do you think?

Link to comment: https://github.com/idx-academy/spa-orders/pull/258#discussion_r1675404438

code/src/pages/cart/CartPage.test.tsx
dmtrth25: > Looks much better, thanks! What about mine test? Should I create separate issue for it now or we can fix it in connection to some PR?

When you'll work with another issue. Don't forget to change. That's not urgent

code/src/pages/cart/CartPage.tsx
dmtrth25: > This can be refactored like I wrote for @D1nviz's PR to avoid `|| {}` and `userId!`. Seems like he is not available for now. Maybe you will merge it faster, so you can also fix it. What do you think?
> 
> Link to comment: [#258 (comment)](https://github.com/idx-academy/spa-orders/pull/258#discussion_r1675404438)

He used `eslint disable`. In my case it should be baecause you know about marker `!` -> `!null && !undefined`
Decided to invistigate this refactor futher.
For me it looks pretty much simplier

mhevyk: Consider following example:

```jsx
 const { id: userId } = useUserDetailsSelector() || {};
```

You need `|| {}` part here because `useUserDetailsSelector`'s return type is `UserDetails | null`. It is related only to typescript part. You will not need those if `useUserDetailsSelector()` will never return `null`. Also it will not be required to do in other places through the code, that will drastically cleans up our codebase

code/src/pages/cart/CartPage.test.tsx
mhevyk: I created new issue and self-assigned to it in order to definitely avoid forgetting

code/src/pages/cart/CartPage.tsx
dmtrth25: > Consider following example:
> 
> ```js
>  const { id: userId } = useUserDetailsSelector() || {};
> ```
> 
> You need `|| {}` part here because `useUserDetailsSelector`'s return type is `UserDetails | null`. It is related only to typescript part. You will not need those if `useUserDetailsSelector()` will never return `null`. Also it will not be required to do in other places through the code, that will drastically cleans up our codebase

As I talked. When we discussed with @D1nviz . For refactor should be another issue. We should move forward with tasks and not be busy with refactoring

mhevyk: We don't need to check for authorization twice, because it is already performed by `ProtectedRoute` component. We should not ever call `useUserDetailsSelector` when we are not under `ProtectedRoute`. Do you agree with it?

mhevyk (approved):

OliaHrytsak (approved):

code/src/types/user.types.ts
```diff
@@ -23,3 +23,7 @@ export type TokenPayload = {
 };
 
 export type UserDetails = User & TokenPayload;
+
+export type UserId = {
+  userId: number;
+};
```
D1nviz, line 31: also use this type for BaseUser. Like on the screenshot
image

dmtrth25: > also use this type for BaseUser. Like on the screenshot image

There will be an error I guess because `userId` and `id` are different
image

D1nviz: Yes. Because you did UserId like an object. It's just a number . Make it like on the screenshot and everything will be okay. 
`export type UserId = number`.

D1nviz (approved):

PR #274, opened 2024-07-12
Added real cart items counter to header cart icon

### LINKED ISSUE #268: Display cart items count inside of cart icon in header
Add real cart items count on icon in header

heshvab:
- Added request for cart items. As you can see, I used lazy query that is initiated from `useEffect`, I did it because the user can be unauthorized - in such case there is no point in making request.
- If user has 0 items or he is unauthorized - badge is not displayed. If the user has more than 99 items - `99+` is displayed. In other cases quantity of items in a cart is displayed.
- As for tests, I think `HeaderToolbar` should be decomposed so I am going to make an issue for it that will include missing tests.

Examples
image

image

image

image


code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -71,10 +74,37 @@ const HeaderToolbar = () => {
...
+  useEffect(() => {
+    if (user?.id && isAuthenticated) {
+      fetch(user.id);
+    }
+  }, [user?.id]);
+
+  const cartItemsCount = data?.items.length;
+
+  const displayedCartItemsCount =
+    cartItemsCount && cartItemsCount > 99 ? 99 + "+" : cartItemsCount;
```
OliaHrytsak: I think you can use "99+" here, instead of 99+"+"

heorhiiss: Thank you, fixed

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -71,10 +74,37 @@ const HeaderToolbar = () => {
...
 
-  const badgeContentTypography = (
-    <AppTypography variant="caption-small">{itemsInCartCount}</AppTypography>
+  useEffect(() => {
+    if (user?.id && isAuthenticated) {
+      fetch(user.id);
+    }
+  }, [user?.id]);
+
+  const cartItemsCount = data?.items.length;
```
OliaHrytsak: Now the items counting isn't correct, because you use items.length but not quantity
I offer to change in this way
`const cartItemsCount = data?.items.reduce((total, item) => total + item.quantity, 0)`
It is how cart.data looks in console
image
This view with changes
image

heorhiiss: Thank you for suggestion. I initially thought to not include quantity and count item as one always, even if its quantity is bigger than one (somehow it was logical to me). But your suggestion is the right one, I checked other e-commerce websites, and they have it the same way as you proposed. Fixed

OliaHrytsak (approved):

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -41,6 +43,10 @@ const HeaderToolbar = () => {
   const userRole = useUserRoleSelector();
   const dispatch = useAppDispatch();
 
+  const [fetch, { data }] = useLazyGetCartItemsQuery();
```
mhevyk: I suggest using another name to avoid collisions with native javascript `fetch`

heorhiiss: Thank you. Changed it to `fetchCartItems`

D1nviz (approved):

mhevyk (approved):

dmtrth25 (approved):

PR #280, opened 2024-07-14
Changed product card "add to cart" logic

### LINKED ISSUE #257: Change logic after 'add to cart' button click
Remove logic that opens cart drawer when user clicks `add to cart`. Instead of this, we can show some kind of snackbar or just show state on `ProductCard`. Drawer must appear only when user clicks cart button in the right corner of app (already implemented).

heshvab:
- Removed drawer opening on add to cart click
- Added logic to display current product state (whether it was added to cart). For now it is done of client side
- Replaced AddToCart button with icon
- Added logic for mock-backend so we can add and remove products from cart
- Gonna add tests later

UI
image

For authorized users

https://github.com/user-attachments/assets/08f5ae9d-4610-4fb6-a8da-2eece442c63d

For not authorized

https://github.com/user-attachments/assets/dbf14493-09de-40f6-805a-c8edb9ab9c28


OliaHrytsak (approved):

code/src/assets/icons/CartOutlined.tsx
```diff
@@ -0,0 +1,51 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+    xmlnsXlink="http://www.w3.org/1999/xlink"
+    height="28px"
+    width="28px"
+    viewBox="0 0 12 12"
+    xmlSpace="preserve"
+    {...props}
+  >
+    <g>
+      <path
```
mhevyk: Using `svg` as component is generally considered bad approach because it's code goes to javascript bundle. We should try to store it inside assets or use package for it, for example for `vite` it is `vite-plugin-svgr`, I think there should be package for `webpack` also. As far as I can see, these components do not depend from props, so we can move it to assets.

Recently I came up with interesting video and I strongly recommend to watch it. Here is the link: https://www.youtube.com/watch?v=5r9sN5Yz79A

I hope it will be helpful for you

code/src/containers/products-container/ProductsContainer.tsx
```diff
@@ -44,31 +79,47 @@ const ProductsContainer = ({
...
+          messageTranslationKey: "cart.itemAddition.fail"
+        });
       }
-    } catch {
-      openSnackbarWithTimeout({
-        variant: "error",
-        messageTranslationKey: "cart.itemAddition.fail"
-      });
+    } else {
+      openModal(<AuthModal />);
```
mhevyk: Try to use guard clause here. In my point of view it will make our code more readable

Here is an article to recollect this concept: https://medium.com/@timothydan/javascript-guard-clauses-64b999e3240

heorhiiss: Thank you, fixed

code/src/assets/icons/CartOutlined.tsx
heorhiiss: Thank you for suggestion and for video, I found out new information for myself. I changed it in favour of the approach shown in video.

D1nviz (approved):

code/src/components/product-card/ProductCard.tsx
```diff
@@ -41,10 +64,17 @@ const ProductCard = ({ product, onAddToCart }: ProductCardProps) => {
...
         </AppTypography>
-        <AppButton size="small" variant="shadow" onClick={handleAddToCart}>
-          <AddIcon />
-          <AppTypography translationKey="productCard.add" />
-        </AppButton>
+        <AppIconButton
+          onClick={handleCartIconClick}
+          className={cn(
+            "spa-product-card__cart-button",
+            isProductInCart && "spa-product-card__cart-button_active"
```
mhevyk: I suggest using modifier here, like `spa-product-card__cart-button--active`

mhevyk (approved):

code/src/assets/icons/CartOutlined.tsx
mhevyk: You are welcome. I am happy to hear that it was useful for you

code/src/components/product-card/ProductCard.tsx
heorhiiss: Thank you. I guess there are two approaches for adding modifiers because `_` was used in documentation I read. At the same time, in other tutorials and documentation `--` was used. I changed it as you proposed to keep our codebase consistent.

PR #284, opened 2024-07-14
Add AppRangeSlider component

### LINKED ISSUE #283: Filter component: AppRangeSlider

mhevyk:
Created AppRangeSlider component for future price filtering

https://github.com/user-attachments/assets/d8f0600a-68f4-467a-89a6-e6d9e82c555b

Inspired by implementation on Rozetka:

image


D1nviz (approved):

OliaHrytsak (approved):

heshvab (approved):

code/src/components/app-range-slider/AppRangeSlider.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+import AppRangeSlider from "@/components/app-range-slider/AppRangeSlider";
+
+import typeIntoInput from "@/utils/type-into-input/typeIntoInput";
+
+const mockHandleChange = jest.fn();
+
+describe("AppRangeSlider", () => {
+  describe("with default props", () => {
+    test("uses default props correctly", () => {
+      render(<AppRangeSlider />);
```
dmtrth25: use `renderComponent` own utility with `props` as you created early with some other tests
And use it once for other places with `renderWithProviders` instead `render`. Follow single approach for render component

code/src/components/app-range-slider/AppRangeSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+  className,
+  onChange,
+  min = 0,
+  max = 20000,
+  step = 10,
+  value,
+  ...props
+}: AppRangeSliderProps) => {
+  const [rangeStart, setRangeStart] = useState(value?.[0] ?? min);
+  const [rangeEnd, setRangeEnd] = useState(value?.[1] ?? max);
```
dmtrth25: should define variables for state. The state are not hardcoded by anyone 
If you need to change something, just change the variable

code/src/components/app-range-slider/AppRangeSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const commonRangeProps = useMemo(
+    () => ({ min, max, step }),
+    [min, max, step]
+  );
```
dmtrth25: how this `useMemo` memoization works? If we not defined even dependencies.
Can record a little video before and after usage. Just interesting how it works here

code/src/components/app-range-slider/AppRangeSlider.test.tsx
mhevyk: Thanks, fixed

code/src/components/app-range-slider/AppRangeSlider.tsx
mhevyk: Created `initialRangeStart` and `initialRangeEnd` variables, the code really become more readable

code/src/components/app-range-slider/AppRangeSlider.tsx
mhevyk: I don't see any sense of recording a little video for it. I used `useMemo` here because `rangeStart`, `rangeEnd` and `value` states change often, but `min`, `max` ans `step` props are supposed to change rarely, most likely on initial setup, so I tried to save extra performance here

dmtrth25: > I don't see any sense of recording a little video for it. I used `useMemo` here because `rangeStart`, `rangeEnd` and `value` states change often, but `min`, `max` ans `step` props are supposed to change rarely, most likely on initial setup, so I tried to save extra performance here

I just want to see the differences using of it.
Whats happen if we not use memoization ? Please explain it in video. but for me it's not understandable

code/src/components/app-range-slider/AppRangeSlider.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onChange?.(typedEvent, [rangeStart, limitedNewValue]);
+  };
+
+  const handleSliderChange = (event: Event, newValue: number | number[]) => {
+    const typedNewValue = newValue as number[];
+
+    setRangeStart(typedNewValue[0]);
+    setRangeEnd(typedNewValue[1]);
+    onChange?.(event, typedNewValue);
+  };
```
dmtrth25:
```
const value = event.target.value;
const numericValue = parseInt(value);
const limitedNewValue = Math.max(min, numericValue);
```
 are repeated please improve it
 
 Maybe should to invistigate and create separate function for both places

code/src/components/app-range-slider/AppRangeSlider.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      renderWithProviders(
+        <AppRangeSlider
+          value={[10, 40]}
+          min={0}
+          max={100}
+          step={10}
+          onChange={mockHandleChange}
+        />
+      );
+    });
```
dmtrth25, line 33: `renderWithProviders` repeated for now. As I told create own render function
You did it before and call inside of tests

image

code/src/components/app-range-slider/AppRangeSlider.tsx
mhevyk: I brainstormed that we can remove it for now, because in this case it is not a big deal

code/src/components/app-range-slider/AppRangeSlider.types.ts
```diff
@@ -0,0 +1,27 @@
+import { ChangeEvent } from "react";
+
+import { SliderProps } from "@mui/material/Slider";
+
+type AppRangeClassNames = {
+  root: string;
+  range: string;
+  toolbar: string;
+  toolbarInput: string;
+};
```
dmtrth25: what do you think.
If all values of keys are `string`
can we improve it ?

code/src/components/app-range-slider/AppRangeSlider.test.tsx
mhevyk: I tried to do it, but it does not make much sense because props differ very much. A intentionally created two `describe` blocks to take advantage of `beforeEach` hook for solving code reusability problem

image

code/src/components/app-range-slider/AppRangeSlider.types.ts
mhevyk: Refactored as follows:

```tsx
type AppRangeClassName = "root" | "range" | "toolbar" | "toolbarInput";
type AppRangeClassNames = Partial<Record<AppRangeClassName, string>>;
```

code/src/components/app-range-slider/AppRangeSlider.tsx
mhevyk: Created helper function:

```tsx
const parseRangePositionValue = (
  event: ChangeEvent<HTMLInputElement>,
  limiterFn: typeof Math.max | typeof Math.min,
  edgeValue: number
) => {
  const value = event.target.value;
  const numericValue = parseInt(value);
  const limitedNewValue = limiterFn(edgeValue, numericValue);

  // needed for typescript, we don't need to pass first parameter at all
  const typedEvent = event as unknown as Event;

  return { typedEvent, limitedNewValue };
};
```

code/src/components/app-range-slider/AppRangeSlider.test.tsx
dmtrth25: > I tried to do it, but it does not make much sense because props differ very much. A intentionally created two `describe` blocks to take advantage of `beforeEach` hook for solving code reusability problem
> 
> image

Ok. but try to write with your single approach with `renderComponent` or not. Just differences with your style code logic in separate cases

dmtrth25 (approved):

code/src/components/app-range-slider/AppRangeSlider.test.tsx
mhevyk: I agree with you, but in different cases we should pick the best approach to solve the problem. You wouldn't use a screwdriver to hammer a nail, the same is true here

PR #289, opened 2024-07-15
Added logic to get items via drower and to remove on both places

### LINKED ISSUE #262: Include cart items in Drawer
When the user add product to the cart the items also should display in the Drawer

### LINKED ISSUE #282: Add delete item logic to cart page
Add possibility for user to delete item from a cart on cart page

dmtrth25:
Added logic to `GET` items according to `user` via `Drawer` and possibility to remove items on the `Cart` page and on the `Drawer`

https://github.com/user-attachments/assets/233c0ad4-703b-49a3-9e69-8c7533a6be76

- created custom hook `useUserCartItems` which uses in both places for `CartPage` and for `CartDrawer`
- some `length` check condition. hope it'll be improved further
- refactored some types

**Resolved: #262 #282**


code/src/types/cart.types.ts
```diff
@@ -27,3 +27,10 @@ export type CartType = {
   items: CartItem[];
   totalPrice: number;
 };
+
+export type CartItemProps = {
+  item: CartItem;
+  onRemove: (product: CartItem) => void;
+};
```
D1nviz, line 34: move this type to file with this Component. We are not defining the types for the components props inside global types

code/src/types/cart.types.ts
```diff
@@ -27,3 +27,10 @@ export type CartType = {
...
+
+export type CartDrawerItemProps = Pick<CartItemProps, "onRemove"> & CartItem;
```
D1nviz, line 36: also here. Just move it to the file where this component is placed, because it's not a global type, not a separate entity, it is just props for the component.

dmtrth25: > also here. Just move it to the file where this component is placed, because it's not a global type, not a separate entity, it is just props for the component.

Also it depends with `CartItemProps` I wrote above. What do you think ?

code/src/types/cart.types.ts
dmtrth25: > move this type to file with this Component. We are not defining the types for the components props inside global types

In this case I tried to move it inside component. But I repeated it twice in `CartItem` and `CartDrawerItem`.
Suggest to leave it there as global common type

heshvab (approved):

code/src/containers/cart-drawer/CartDrawer.tsx
```diff
@@ -1,54 +1,54 @@
...
-      onRemove={handleRemoveItem}
-      {...item}
-    />
-  ));
+  const renderCartItems =
+    cartItemsList.length > 0 ? (
+      cartItemsList
+    ) : (
+      <AppTypography variant="subtitle2" translationKey="cart.emptyItem" />
+    );
```
mhevyk: I suggest to rename it to `cartItemsContent` because with name `renderCartItems` seems like function to me

mhevyk (approved):

D1nviz (approved):

PR #301, opened 2024-07-16
Add filtration for orders on admin dashboard

### LINKED ISSUE #256: Add filtration for orders
Open api spec
https://github.com/idx-academy/mic-orders/blob/feature-126-orders-management-page/code/orders-api-rest/api/openapi-rest.yml

image

mhevyk:
https://github.com/user-attachments/assets/0944b4b5-7063-4cbf-901c-fa8b8ee8470a

- [x] Added reusable `useFiltersWithApply` hook

It handles filters in state and saves them into params when we apply them. Also this hook stores active filters. Active filter is a filter that is changed in some way and we should reset it manually. It resembles form validation when form becomes dirty we need to reset in manually. I did it because it is impossible to handle resetting to initial state correctly because initial state can be either `undefined`, `null`, `false`, `0` and so on.

I created custom serializers and deserializers for search params to save to params and extract from them already correct type to avoid casting it manually every time we get param from url. I added support for `number`, `string`, `boolean`, `Set` and `FilterRange`

Why I used `Set`? Because I think it is very efficient in storing multiple options, like checkboxes and it can easily be converted to array when we apply filters using `Array.from()`.

Set `new Set([1, 2, 3])` serializes into `{1,2,3}` and same for deserialize. I used regex for it for more exact parsing

What is `FilterRange` type?

It is just object with start and end properties like follows:

```tsx
const price: FilterRange = {
    start: 10,
    end: 30
}
```

This object can be serialized into 10-20, example in url `http://localhost:3000/dashboard?price=10-20`

- [x] Added button, that opens filter drawer
- [x] Created endpoint for mock backend and added one more mock orders
- [x] Created rtk types and changed queries to use filters correctly
- [x] Added time span filter. We can filters orders performed yesterday for example, last week, last two weeks and last month. It the future we can add more options, but i think it is enough for now

Time span is stored as string and it delegates to function `timeSpanToDateRange` that returns correct filter range

- [x] Created `AppDrawer` component. This task required this change because using existing drawer context did not work for me in any way because dynamic state updated from parent did not affect drawer in any way. We should try to use drawer context if we can and if we can't, like in my case, we can use `AppDrawer`.
- [x] Changed some app components for styling or reusability purposes

I will write tests in separate issue #341


code/src/containers/dashboard-tabs/components/orders-tab-filter-drawer/OrdersTabFilterDrawer.tsx
```diff
@@ -1,36 +1,137 @@
...
-import { useDrawerContext } from "@/context/drawer/DrawerContext";
+import { timeSpans } from "@/constants/timeSpans";
+import { FilterActions } from "@/hooks/use-filters-with-apply/useFiltersWithApply.types";
+import { ExtractSetValue, TimeSpan } from "@/types/common";
+import { DeliveryMethod } from "@/types/delivery.types";
+import { OrderStatus } from "@/types/order.types";
 
 import "@/containers/dashboard-tabs/components/orders-tab-filter-drawer/OrdersTabFilterDrawer.scss";
 
+import { AdminOrderFilters } from "../../hooks/use-filtered-admin-orders/useFilteredAdminOrders.types";
```
D1nviz: use alias

code/src/containers/dashboard-tabs/components/orders-tab-filter-drawer/OrdersTabFilterDrawer.tsx
```diff
@@ -1,36 +1,137 @@
...
+      labelId={selectLabelId}
+    >
+      <AppMenuItem value="" disabled>
+        <AppTypography translationKey="select.defaultOption" />
+      </AppMenuItem>
+      {Object.entries(timeSpans).map(([datePeriod, translationKey]) => (
+        <AppMenuItem value={datePeriod} key={datePeriod}>
+          <AppTypography translationKey={translationKey} />
+        </AppMenuItem>
+      ))}
```
D1nviz: I think it would be better to move it above the return statement

D1nviz:
Maybe create folder inside the utils for these helpers
image

D1nviz:
Great job!

code/src/containers/dashboard-tabs/components/orders-tab-filter-drawer/OrdersTabFilterDrawer.tsx
mhevyk: Thank you! Moved

code/src/containers/dashboard-tabs/components/orders-tab-filter-drawer/OrdersTabFilterDrawer.tsx
mhevyk: Thanks for advice! I did it!

mhevyk:
> Maybe create folder inside the utils for these helpers image

I didn't want to put it inside utils, because these are local functions that are used only in one place. I split it into different files to make it easier to test and to make the file with the hook smaller

mhevyk:
> Great job!

Thanks!

D1nviz (approved):

heshvab (approved):

heshvab:
Nice job!

mhevyk:
> Nice job!

I'm incredibly grateful!

OliaHrytsak (approved):

PR #343, opened 2024-07-19
Added test for i18nProvider

### LINKED ISSUE #337: Add test for i18n context

D1nviz:
image


OliaHrytsak (approved):

code/src/context/i18n/I18nProvider.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    test("sets the locale correctly based on browser locale", () => {
+      Object.defineProperty(navigator, "language", {
+        value: "uk-UA",
+        configurable: true
+      });
+
+      renderWithProviders(<I18nPlayground />);
+
+      const localeDisplay = screen.getByTestId("localeDisplay");
```
dmtrth25: what do you think if we use `getByTestId("localeDisplay")` on the `beforeEach` block can we remove this ?
And also look at the other tests where it's present. I guess it can be removed

code/src/context/i18n/I18nProvider.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      jest.restoreAllMocks();
+    });
+
+    test("sets the locale correctly based on browser locale", () => {
+      Object.defineProperty(navigator, "language", {
+        value: "uk-UA",
+        configurable: true
+      });
+
+      renderWithProviders(<I18nPlayground />);
```
dmtrth25, line 82: also can remove this
It used in `beforeEach`

code/src/context/i18n/I18nProvider.test.tsx
D1nviz: We cannot move it to before each because these test cases differ. As you can see for some tests I must mock and render, for some I don't need this option. Also, for some cases, I need to define the properties of the navigator object, so I don't know how to use beforeEach here, they are different

image

dmtrth25: > We cannot move it to before each because these test cases differ. As you can see for some tests I must mock and render, for some I don't need this option. Also, for some cases, I need to define the properties of the navigator object, so I don't know how to use beforeEach here, they are different
> 
> image

Look. I'm not talk about moving it into `beforeEach` etc.
You are already has this on the `beforeEach`
image

Can remove that variable. I can show other your cases without using it just expect
image

I mean that can remove the variable here

dmtrth25 (approved):

mhevyk (approved):

mhevyk (approved):
Great work! Interesting playground solution :)

PR #347, opened 2024-07-20
Increased containers coverage

### LINKED ISSUE #319: Increase coverage for containers
image
image
image

image
image
image

image
image
image

heshvab:
Mutation coverage
image

Istanbul coverage
image

I haven't tested components inside of `dashboard-tabs`, because logic related to filters there is still in process of developing.

There is a mutant like this I can't cover. I also haven't found a way to not render Icon in a different way.
image


code/src/containers/language-select/LanguageSelect.test.tsx
```diff
@@ -1,17 +1,23 @@
-import { act, fireEvent, screen } from "@testing-library/react";
+import { act, fireEvent, screen, waitFor } from "@testing-library/react";
 
 import LanguageSelect from "@/containers/language-select/LanguageSelect";
 
+import { LOCAL_STORAGE_KEYS } from "@/constants/common";
 import locales from "@/constants/locales";
 import renderWithProviders from "@/utils/render-with-providers/renderWithProviders";
 
+jest.spyOn(window.localStorage.__proto__, "setItem");
```
mhevyk: `jest.spyOn(window.localStorage, "setItem")` does not work? As alternative we can use `spyOn(Storage.prototype, 'setItem')`, it look shorter:

https://stackoverflow.com/questions/11485420/how-to-mock-localstorage-in-javascript-unit-tests

code/src/containers/forms/sign-up-form/SignUpForm.test.tsx
```diff
@@ -25,6 +26,16 @@ const mockCloseModal = jest.fn();
...
 
+jest.mock("@/hooks/use-input-visibility/useInputVisibility", () => ({
+  __esModule: true,
+  default: jest
+    .fn()
+    .mockImplementation(
+      jest.requireActual("@/hooks/use-input-visibility/useInputVisibility")
+        .default
+    )
+}));
```
mhevyk, line 38: Why we need so complex mock here? just `jest.fn()` won't be enough?

OliaHrytsak (approved):

code/src/containers/language-select/LanguageSelect.test.tsx
heorhiiss: Thank you. For some reason, `jest.spyOn(window.localStorage, "setItem")` doesn't work. I changed that line with suggested `spyOn(Storage.prototype, 'setItem')`.

code/src/containers/forms/sign-up-form/SignUpForm.test.tsx
heorhiiss: I needed to make sure that one of the imports was called and at the same time not change its implementation. That is the only way I found that works well. In simpler case `jest.spy` could be used, but it doesn't work in this case, maybe because there are restrictions when using it with imports.

D1nviz (approved):

code/src/containers/forms/sign-up-form/SignUpForm.test.tsx
mhevyk: Ok, i got it! Unconventional solution! But I recommend adding a comment to it to avoid unpurposly deleting by other developer

heorhiiss: Added a comment as you suggested.

mhevyk (approved):

PR #354, opened 2024-07-22
Tests for useFiltersWithApply and related utils

### LINKED ISSUE #341: Add tests for filter-related functionality (part one)
Add tests for useFiltersWithApply and all related utils

mhevyk:
Next part of tests for filter functionalities will be added in #355

Unit tests:
image
image

Mutation tests:
image
image


D1nviz (approved):

code/src/utils/time-span-to-date-range/timeSpanToDateRange.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
+
+  test.each(testCases)(
+    "returns correct date range for %s",
+    (timeSpan, expectedStart) => {
+      const { start, end } = timeSpanToDateRange(timeSpan as TimeSpan);
+      console.log({ start, end });
```
D1nviz: remove `console.log`

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  "@/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams"
+);
+jest.mock(
+  "@/hooks/use-filters-with-apply/serialize-to-query-string/serializeToQueryString"
+);
+
+const mockSetSearchParams = jest.fn();
+
+describe("useFiltersWithApply", () => {
+  let defaultFilters: Record<string, unknown>;
```
heorhiiss: Is there a need to create this variable here and than set value in `beforeEach`? Maybe create it as const outside of describe with the value that setted in `beforeEach`?

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
```diff
@@ -0,0 +1,29 @@
+import parseFiltersFromParams from "@/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams";
+
+describe("parseFiltersFromParams", () => {
+  test("returns default filters when no filters are applied inside params", () => {
+    const defaultFilters = { filter1: "value1" };
```
dmtrth25: move mocked data our of the `describe` block.
Maybe also with `parsedValue`

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    );
+
+    expect(parsedValue.defaultActiveFilters.size).toBe(0);
+    expect(parsedValue.defaultFiltersFromParams).toEqual(defaultFilters);
+  });
+
+  test("prefers using filter from params to default filter", () => {
+    const defaultFilters = { filter2: "value1" };
+    const searchParams = new URLSearchParams({ filter2: "value2" });
+    const parsedValue = parseFiltersFromParams(defaultFilters, searchParams);
```
dmtrth25: can we create it above ? Or reuse existing mocked data ? with `filter1` e.g. I mean `defaultFilters`

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    expect(parsedValue.defaultActiveFilters.size).toBe(1);
+    expect(parsedValue.defaultFiltersFromParams).toEqual({ filter2: "value2" });
```
dmtrth25: in example above you use `toEqual(variable)`
So. As I see it's doesn't matter value because it's mocked. Reuse once
It's not used `defaultFilters` variable here in this test but it was defined

code/src/hooks/use-filters-with-apply/parse-serialized-range/parseSerializedRange.test.ts
```diff
@@ -0,0 +1,20 @@
+import parseSerializedRange from "@/hooks/use-filters-with-apply/parse-serialized-range/parseSerializedRange";
+
+describe("parseSerializedRange", () => {
+  test("parses range of two integers correctly", () => {
+    const expectedRange = { start: 10, end: 20 };
```
dmtrth25: mocked data. move above. check in other places

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.test.ts
```diff
@@ -0,0 +1,30 @@
+import parseSerializedSet from "@/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet";
+
+describe("parseSerializedSet", () => {
+  test("parses correct serialized set correctly", () => {
+    const serializedSet = "{completed,canceled,delivered}";
+    const expectedDeserializedSet = new Set([
+      "completed",
+      "canceled",
+      "delivered"
+    ]);
```
dmtrth25: also mocked data. Check other places

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.ts
```diff
@@ -1,23 +1,27 @@
...
+const parseSerializedSet = <Value>(value: string) => {
+  const match = value.match(serializedSetPattern);
 
   const arrayMatch = match?.[1];
 
-  if (arrayMatch === "") {
-    return new Set();
+  if (arrayMatch === undefined) {
+    return null;
   }
```
dmtrth25, line 14: I have suggestion
```suggestion
if (!arrayMatch) return null
```

code/src/hooks/use-filters-with-apply/serialize-to-query-string/serializeToQueryString.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  test("serializes 'boolean' type correctly", () => {
+    expect(serializeToQueryString(false)).toBe("false");
+    expect(serializeToQueryString(true)).toBe("true");
+  });
+
+  test("serializes 'Set' type correctly", () => {
+    const serializationResult = serializeToQueryString(
+      new Set(["value1", "value2"])
+    );
```
dmtrth25: mocked. Also in other places

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    ]);
+
+    (parseFiltersFromParams as jest.Mock).mockReturnValue({
+      defaultActiveFilters: new Set(),
+      defaultFiltersFromParams: defaultFilters
+    });
+
+    (serializeToQueryString as jest.Mock).mockImplementation((value: object) =>
+      value.toString()
+    );
```
dmtrth25: move `mocks` above.
Also will be better create like

image

Can you imrpove it ? Maybe some general. But `jest.Mock.mockReturnValue` repeated once at least
What do you think ?

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      result.current.actions.updateFilterByKey("filter2", "updated2");
+      result.current.actions.resetFilters();
+    });
+
+    expect(result.current.filters).toEqual(defaultFilters);
+    expect(result.current.activeFiltersCount).toBe(0);
+  });
+
+  test("applies filters correctly", () => {
+    const { result } = renderHook(() => useFiltersWithApply(defaultFilters));
```
dmtrth25: check it uses it each test. Can improve it

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
dmtrth25: maybe also define with `const` once be better. Because this variable has rewritten only once in `beforeEach` ?

code/src/utils/time-span-to-date-range/timeSpanToDateRange.test.ts
mhevyk: My bad! Removed

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
mhevyk: Thanks! Moved `defaultFilters` to constant variable outside describe block as follows:

```tsx
const defaultFilters = { filter1: "default1", filter2: "default2" };
```

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
mhevyk: Thanks a lot! Added two mock variables outside describe and reused them

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
mhevyk: Yeah, we can. I did as you suggested

code/src/hooks/use-filters-with-apply/parse-filters-from-params/parseFiltersFromParams.test.ts
mhevyk: I kept `toEqual`, but reused mock variable

code/src/hooks/use-filters-with-apply/parse-serialized-range/parseSerializedRange.test.ts
mhevyk: Thanks! I created two constants as follows:

```tsx
const integerRange = { start: 10, end: 20 };
const decimalRange = { start: 10.25, end: 30.5 };
```

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.test.ts
mhevyk: Thanks! Created variable:

```tsx
const setWithStatuses = new Set(["completed", "canceled", "delivered"]);
```

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.ts
mhevyk: Thanks for suggestion! Maybe it is not required here. I like to have brackets because we can add more content to if block in less steps and for me it looks more readable. But if you insist I can apply your suggestion

heshvab (approved):

code/src/hooks/use-filters-with-apply/serialize-to-query-string/serializeToQueryString.test.ts
mhevyk: Thanks, created mocked variable for this case

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
mhevyk: Used your suggested approach, thank you Dmytro

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
mhevyk: Tried ny best to improve it. Please rereview

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.test.ts
dmtrth25: > Thanks! Created variable:
> 
> ```tsx
> const setWithStatuses = new Set(["completed", "canceled", "delivered"]);
> ```

`const serializedSet = "{completed,canceled,delivered}";` is also mocked

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.ts
dmtrth25: > Thanks for suggestion! Maybe it is not required here. I like to have brackets because we can add more content to if block in less steps and for me it looks more readable. But if you insist I can apply your suggestion

If you want with `{}` ok.
But replace check with `!arrayMatch` compare with checking with `undefined` it's not a good solution

code/src/hooks/use-filters-with-apply/useFiltersWithApply.test.ts
dmtrth25: > Tried ny best to improve it. Please rereview

I just saw it's repeated also in `renderAndMock`
We should move on and not concentrate for that I guess

code/src/hooks/use-filters-with-apply/deserialize-from-query-string/deserializeFromQueryString.ts
```diff
@@ -2,7 +2,7 @@ import parseSerializedRange from "@/hooks/use-filters-with-apply/parse-serialize
 import parseSerializedSet from "@/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet";
 
 const checkNumber = (value: unknown) => {
-  return Number(value) === value;
+  return !isNaN(value as number);
 };
```
dmtrth25: suggestion to avoid `unknown`
```suggestion
const checkNumber = (value: string) => {
  return !isNaN(Number(value));
};
```

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.ts
mhevyk: Sorry, I can't do it. I hope you know how [`match()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) works. For me it has strange interface, but matching against regex is very powerfull.

I want to explain why i did it like it is. `match` works so that if there is no match, it returns empty matched array, so `arrayMatch` will always be `undefined` if we try to access `array[1]` when `array` is empty. If there is match, array of matches is returned. Match are things wrapped inside brackets. For example `(match1, (match2, (match3)))` with return array of `[(match1, (match2, (match3))), (match2, (match3)), (match3)]` if there is a match against regex. We also can ignore matching some group using non-capturing group operator, like i did in regex above. [Read more about non-capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Non-capturing_group)

In my specific regex i can parse empty set from string `'{}'`. The second match in my case will be emty string, so i can check it below and return empty set. Thats why I can't use `!arrayMatch` because as you know it accepts all falsy values, in my case both `""` and `undefined`, but it is not what I want here

code/src/hooks/use-filters-with-apply/deserialize-from-query-string/deserializeFromQueryString.ts
mhevyk: I'm incredibly grateful for really useful suggestion. I applied it, looks much better now!

code/src/hooks/use-filters-with-apply/parse-serialized-set/parseSerializedSet.test.ts
mhevyk: Thanks, I also moved it above `describe` block

dmtrth25 (approved):

PR #358, opened 2024-07-23
Created util function for filtering products by category

### LINKED ISSUE #329: Create logic to change category of products
For each category should be separate page or single page with query params for search (different content but the same layout)

image

ValKoDim:
Created a util function that filters list of products by their category. The function accepts an array of Product objects and a category string. I have also created a test for this new function.


mhevyk (commented):
image

Please change title when we change filter. Also it wouldbe good if you could change label called `6 products` to `6 tablets`, `6 mobiles`. Hope you got my thought

code/src/pages/products/ProductsPage.test.tsx
```diff
@@ -41,7 +41,7 @@ jest.mock("@/store/api/productsApi", () => ({
 }));
 
 const defaultQueryArguments = {
-  size: 10,
+  size: Infinity,
```
mhevyk: Why we need `Infinity` here?

code/src/pages/products/ProductsPage.tsx
```diff
@@ -29,10 +30,14 @@ const ProductsPage = () => {
     isError
   } = useGetProductsQuery({
     page: page - 1,
-    size: 10,
+    size: Infinity,
```
mhevyk: Same question

code/src/utils/filter-products-by-category/categoryFilter.ts
```diff
@@ -0,0 +1,15 @@
+import { Product } from "@/types/product.types";
+
+const categoryFilter = (
+  category: string | null,
+  productsList: Product[] | undefined
```
mhevyk: I suggest adding comment for line `Product[] | undefined` that explains why we use `productsList: Product[] | undefined` instead of `productsList?: Product[]`

code/src/pages/products/ProductsPage.test.tsx
ValKoDim: This is just a temporary solution. It is made so, i can pull list of all products from db, because currently db requires size of the elements that we get from DB. When we implement endless scrolling, then we will go back to using a realistic value for that DB request.

code/src/pages/products/ProductsPage.tsx
ValKoDim: This is just a temporary solution. It is made so, i can pull list of all products from db, because currently db requires size of the elements that we get from DB. When we implement endless scrolling, then we will go back to using a realistic value for that DB request.

code/src/pages/products/ProductsPage.test.tsx
mhevyk: Okay, I got it. So please add comment on the right of this line to avoid confusion among other developers when they work with this code

ValKoDim: Done

code/src/utils/filter-products-by-category/categoryFilter.ts
ValKoDim: I changed to syntax to what you suggested. I agree. that it looks much more readable.

code/src/pages/products/ProductsPage.tsx
```diff
@@ -41,29 +46,28 @@ const ProductsPage = () => {
...
 
   return (
     <PageWrapper>
       <AppBox className="spa-products-page" data-cy="products-page">
         <AppTypography
           variant="h3"
           className="spa-products-page__header"
-          translationKey="productsAll.label"
+          translationKey={
+            !categoryType ? "productsAll.label" : `productsAll.${categoryType}`
```
heorhiiss: Usually we move ternary conditions like this in variable above return statement. It makes code more readable.
Something like this before return statement ```const productsAllLabel = !categoryType ? "productsAll.label" : `productsAll.${categoryType}```
And then in jsx `translationKey={productsAllLabel}`
You can make similar improvement with others ternary conditions you have inside of jsx

ValKoDim: I moved the ternary operators into variables, in order to improve readability.

heshvab (approved):

dmtrth25 (approved):

PR #364, opened 2024-07-23
Aded new product page and restructured admin panel layout

### LINKED ISSUE #357: Create new product route
Create new product route and page (page without content)
I guess it will also require changing current tabs implementation

heshvab:
- Instead of query pamateres, I used simple pages with layout which includes menu from the left side.
- Instead of OrdersTab, ProductsTab and UsersTab added corresponding pages.
- Added DashboardLayout.
- Since tabs became separate pages I also changed structure of admin panel files in our codebase. Layout and its components are located In layouts. I moved orders filter drawer and its components in containers folder. 
- Added new product page


mhevyk (commented):
Visually it looks amazing! I noticed that some disable-like styles are applied to tabs. I suggest removing them

image

code/src/constants/routes.ts
```diff
@@ -21,7 +21,23 @@ const routes = {
...
+      relativePath: "products",
+      new: {
+        path: "/dashboard/products/new",
+        relativePath: "products/new"
+      }
+    },
+    users: {
+      path: "/dashboard/users",
+      relativePath: "users"
+    }
```
mhevyk: Excellent job! I am curious about `relativePath` keys here. I think they can be kinda inferred from path by just joining paths. We can declare only essential part of url and maybe create util to create absolute or relative paths. What do you think about it?

mhevyk (commented):
Also I did not notice tests for dashboard products and dashboard-new-products page. You did not write them because you are working on these tabs, am I correct?

heshvab:
> Also I did not notice tests for dashboard products and dashboard-new-products page. You did not write them because you are working on these tabs, am I correct?

Yeap, that is correct. There are no much sense of writing tests without ready implementation like fetching products

code/src/constants/routes.ts
heorhiiss: Thanks for suggestion. I found out that I don't really need relative paths in this case, so I removed all them

heshvab:
> Visually it looks amazing! I noticed that some disable-like styles are applied to tabs. I suggest removing them
> 
> image

Thanks. It was because AppLink setted some styles when the link is active. Fixed

mhevyk (approved):

D1nviz (approved):

dmtrth25 (approved):

PR #403, opened 2024-07-27
Add useErrorPageRedirect hook

### LINKED ISSUE #393: Add reusable hook or util to redirect to error pages if something went wrong

mhevyk:
This hook can we used for redirect to error pages. For now error pages in our app are `NotFound` and `unknown`, but we can adjust this list later. We can use this hook to perform redirects to these pages in details pages for example when we don't have product id when fetching product by id, we can pass custom error message and other props to render not found page with that message

That way we can avoid typescript errors with `useParams` for example and render correct page when something was not found or any other error occured. On s2s we had to make a lot o optional chains where it is not needed at all

This hook provides following values:

- `state`: can be used only on error page to get data from page that triggered redirect
- `redirect`: can be used to trigger redirect inside handlers or `useEffect`
- `renderRedirectComponent`: can be used to return a navigate component with correct state

### Example of usage:

image


D1nviz (approved):

heshvab (approved):

OliaHrytsak (approved):

OliaHrytsak:
Wery usefull hook!

PR #405, opened 2024-07-27
Restricted visible tabs for admin and manager

### LINKED ISSUE #385: Change tab displayment under different roles
For now we get forbidden for orders and products list on admin dashboard
Under admin role we should see users tab
Under manager role we should see only products and orders tab

image

mhevyk:
https://github.com/user-attachments/assets/4f1f06f6-9659-4d0f-82ba-e40b38b9da12

image

image


code/src/layouts/dashboard-layout/DashboardLayout.tsx
```diff
@@ -1,37 +1,23 @@
...
+    return <DashboardAdminLayout />;
+  }
 
-const DashboardLayout = () => {
-  const tabLabels = dashboardTabs.map((tab) => (
-    <DashboardTab key={tab.name} tab={tab} />
-  ));
+  if (userRole === ROLES.SHOP_MANAGER) {
+    return <DashboardManagerLayout />;
+  }
```
D1nviz: I guess you can use object here and use layouts by Role as a key

dmtrth25: What do you think to add `switch case` here?
I think that in future we can simpler to add some new roles

mhevyk: Thanks! I used `switch case`

mhevyk: Thanks! I used `switch case`! I discussed it with you

D1nviz (approved):

dmtrth25 (approved):

OliaHrytsak (approved):

heshvab (approved):

PR #407, opened 2024-07-28
Increased coverage of HeaderToolbar and refactored it

### LINKED ISSUE #321: Increase coverage for HeaderToolbar and refactor it
image

image
image
image
image
image
image

heshvab:
- Refactored HeaderToolbar. Created a separate component for search input, and refactored code inside of HeaderToolbar without changing its behaviour. Also added one util function and tests for it in the process of refactoring.
- Increased coverage for HeaderToolbar
- Increased coverage for HeaderCategories

Stryker coverage
image

Istanbul coverage (there is 75% coverage for functions because there is one function that does not have implementation yet)
image


code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.tsx
```diff
@@ -0,0 +1,12 @@
+import { CartItem } from "@/types/cart.types";
```
D1nviz: use ts instead of tsx format here. The same for test

OliaHrytsak (approved):

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.tsx
heorhiiss: Thank you for noticing, fixed

D1nviz (approved):

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.test.ts
```diff
@@ -0,0 +1,39 @@
+import { CartItem } from "@/types/cart.types";
+import calculateCartItemsCounter from "@/utils/calculate-cart-items-counter/CalculateCartItemsCounter";
+
+const generateCart = (length: number, increaseQuantity: boolean = false) => {
+  return Array.from({ length }, (_, i) => ({
+    productId: i + 1,
+    quantity: increaseQuantity ? i + 1 : 1
+  })) as unknown as CartItem[];
+};
```
dmtrth25: suggest to avoid some `unknown` just pass the `CartItem` array type for arguments of fn
like 
```jsx
const generateCart = (length: number, increaseQuantity: boolean = false): CartItem[] => {}
```

heorhiiss: I guess I need `unknown` there because `{ productId: number, quantity: number }` does not satisfy `CartItem` type. Just writing type like `(length: number, increaseQuantity: boolean = false): CartItem[]` leads to ts error in return statement. I could write entire cart item object, but there is no much sense since in this function only `quantity` property is needed

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.ts
mhevyk, line 1: Refactor filename because it is different from default export

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const result = calculateCartItemsCounter(cart);
+    expect(result).toBe(99);
+  });
+
+  test("Should return 99+ when cart has 150 items", () => {
+    const cart = generateCart(150);
+
+    const result = calculateCartItemsCounter(cart);
+    expect(result).toBe("99+");
+  });
```
mhevyk: I suggest to use `test.each` here because tests are rather similar

https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -77,21 +63,64 @@ const HeaderToolbar = () => {
...
+    <AppButton
+      onClick={handleOpenAuthModal}
+      key="auth-button"
+      data-cy="auth-button"
+    >
+      <AppTypography component="span" translationKey="signIn.label" />
+    </AppButton>
+  );
+
+  const cartItemsCount = calculateCartItemsCounter(data.items);
```
mhevyk: I think header component became too complicated to understand. Maybe we can create three separate toolbars for different roles? I think it would be easier to maintain, especailly if we will have more roles in the future. Also then we won't need to import unnesessary components or files if they are not needed at all. We can use lazy imports here then

dmtrth25 (approved):

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.ts
heorhiiss: Thanks, fixed

code/src/utils/calculate-cart-items-counter/CalculateCartItemsCounter.test.ts
heorhiiss: Thanks, used `test.each` there

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
heorhiiss: Thanks for suggestion. I divided that component into components according to existing roles. Now it looks much better, because that component was a mess. You also suggested use of lazy imports, but I am not sure about it. It worsens user experience because loaders appear when user lands on a website and when he logs in. It looks not nice.

code/src/layouts/header/components/header-toolbar/header-admin-toolbar/HeaderAdminToolbar.tsx
```diff
@@ -0,0 +1,54 @@
...
+import "@/layouts/header/components/header-toolbar/HeaderToolbar.scss";
+
+const HeaderAdminToolbar = () => {
+  const dispatch = useAppDispatch();
+
+  const handleLogout = () => {
+    dispatch(logout());
+    dispatch(clearLocalCart());
+    dispatch(cartApi.util.resetApiState());
+  };
```
mhevyk: I noticed that the logic of logout is repeated in toolbar components. I suggest creating component `LogoutButton` or custom hook `useLogout`

code/src/layouts/header/components/header-toolbar/header-shop-manager-toolbar/HeaderShopManagerToolbar.tsx
```diff
@@ -0,0 +1,53 @@
...
+        key="dashboard-button"
+      >
+        <AppIconButton
+          to={routes.dashboard.path}
+          component={AppLink}
+          data-testid="header-dashboard-button"
+          data-cy="dashboard-button"
+        >
+          <DashboardCustomizeIcon fontSize="medium" />
+        </AppIconButton>
```
mhevyk: Maybe also create custom component for `DashboardButton`? It would be very easy to maintain if we could compose header toolbar of smaller pieces like lego

code/src/layouts/header/components/header-toolbar/header-cart-button/HeaderCartButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+      badgeContent={badgeContent}
+      variant="contained"
+      size="small"
+      className="header__toolbar-cart-badge"
+      data-testid="header-cart-badge"
+    >
+      <ShoppingCartIcon className="header__toolbar-icon" fontSize="medium" />
+    </AppBadge>
+  ) : (
+    <ShoppingCartIcon className="header__toolbar-icon" fontSize="medium" />
```
mhevyk: Create separate variable for `shoppingCartIcon`

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
```diff
@@ -1,202 +1,59 @@
...
+      headerNavigation = <HeaderAdminToolbar />;
+      break;
+    case ROLES.SHOP_MANAGER:
+      headerNavigation = <HeaderShopManagerToolbar />;
+      break;
+    case ROLES.USER:
+      headerNavigation = <HeaderUserToolbar />;
+      break;
+    default:
+      headerNavigation = <HeaderUnauthorizedUserToolbar />;
```
mhevyk: I suggest moving this function outside a component or even to a local util as I did in `DashboardLayout.tsx`. This util is local to the folder. You can create something similar to this:

```tsx
import DashboardAdminLayout from "@/layouts/dashboard-layout/components/dashboard-admin-tab-layout/DashboardAdminLayout";
import DashboardManagerLayout from "@/layouts/dashboard-layout/components/dashboard-manager-tab-layout/DashboardManagerLayout";

import { ROLES } from "@/constants/common";
import { UserRole } from "@/types/user.types";

const getDasboardLayoutByRole = (role: UserRole | null) => {
  switch (role) {
    case ROLES.ADMIN:
      return <DashboardAdminLayout />;
    case ROLES.SHOP_MANAGER:
      return <DashboardManagerLayout />;
    default:
      return false;
  }
};

export default getDasboardLayoutByRole;
```

Than we can use it similarly to following:

```tsx
const DashboardLayout = () => {
  const userRole = useUserRoleSelector();

  return getLayoutByRole(userRole);
};
```

mhevyk (commented):
You did a huge amount of work but I think it is worth it! Keep going in the same pace!

code/src/layouts/header/components/header-toolbar/HeaderToolbar.tsx
heorhiiss: Thanks, done

code/src/layouts/header/components/header-toolbar/header-cart-button/HeaderCartButton.tsx
heorhiiss: Thanks, done

code/src/layouts/header/components/header-toolbar/header-shop-manager-toolbar/HeaderShopManagerToolbar.tsx
heorhiiss: I created separate files for each button

code/src/layouts/header/components/header-toolbar/header-admin-toolbar/HeaderAdminToolbar.tsx
heorhiiss: I created both hook and component. I guess hook can be useful in other parts of our application

mhevyk (approved):

PR #417, opened 2024-07-30
Create product details page

### LINKED ISSUE #387: Create product details page

mhevyk:
Developed base functionality (ui mostly). I will add logic to buy now button in separate issue. Also I will create issue for changing scrollbar styles

I oriented on discussed with backend side structure of product

https://github.com/user-attachments/assets/654b8226-4f16-4a5b-ae1f-3048b48829d4

image

image


code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.scss
```diff
@@ -0,0 +1,79 @@
...
+  gap: spacing(8) spacing(15);
+
+  &__image-wrapper {
+    flex: 3;
+    width: 100%;
+    min-width: 300px;
+    max-width: 700px;
+  }
+
+  &__image-wrapper img {
```
dmtrth25: define `img {}` styles inside `&__image-wrapper`

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+    const inStockTypography = screen.getByText("productDetailsPage.inStock");
+    expect(inStockTypography).toBeInTheDocument();
+  });
+
+  test("renders delivery methods correctly", () => {
+    renderAndMock({ data: mockProduct });
+
+    const deliveryMethodImages = screen.getAllByAltText(
+      /dashboardTabs.orders.filters.\w+/
```
dmtrth25, line 141: will it work without the `regex` `\w+/` ? also can check other places with `regex`

code/src/pages/product-details/ProductDetailsPage.tsx
```diff
@@ -0,0 +1,34 @@
...
+import AppBox from "@/components/app-box/AppBox";
+
+import useErrorPageRedirect from "@/hooks/use-error-page-redirect/useErrorPageRedirect";
+import { productNotFoundRedirectConfig } from "@/pages/product-details/ProductsDetailsPage.constants";
+import ProductDetailsContainer from "@/pages/product-details/components/product-details-container/ProductDetailsContainer";
+
+import "@/pages/product-details/ProductDetailsPage.scss";
+
+type ProductDetailsPageParams = {
+  productId: string;
```
D1nviz: I suggest moving it to `products.types`

mock-backend/src/routes/productRoutes.js
```diff
@@ -1,13 +1,15 @@
 const express = require("express");
 const {
   getAllProducts,
+  getProductById,
   getAllManagerProducts,
-  createProduct
+  createProduct,
```
dmtrth25, line 6: maybe to create issue to configure `prettier` to avoid this `,` everywhere for mock backend ?

code/src/pages/product-details/ProductDetailsPage.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const ProductDetailsPage = () => {
+  const { productId } = useParams<ProductDetailsPageParams>();
+  const { renderRedirectComponent } = useErrorPageRedirect();
+
+  if (productId === undefined) {
```
D1nviz: maybe use `!productId`

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { productNotFoundRedirectConfig } from "@/pages/product-details/ProductsDetailsPage.constants";
+import { useGetUserProductByIdQuery } from "@/store/api/productsApi";
+import formatPrice from "@/utils/format-price/formatPrice";
+import getCategoryFromTags from "@/utils/get-category-from-tags/getCategoryFromTags";
+import isErrorWithStatus from "@/utils/is-error-with-status/isErrorWithStatus";
+
+import "@/pages/product-details/components/product-details-container/ProductDetailsContainer.scss";
+
+type ProductDetailsContainerProps = {
+  productId: string;
```
D1nviz: Also, I suggest moving it to the products.types` and import here

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+        />
+      }
+    />
+  );
+
+  const productDescription = product.description
+    .split("/")
+    .map((paragraph) => (
+      <AppTypography key={paragraph}>{paragraph}</AppTypography>
+    ));
```
D1nviz, line 70: What do you think about the creating function for that?

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const isNotFoundOnServer = isErrorWithStatus(error) && error.status === 404;
+
+  if (isNotFoundOnServer) {
+    return renderRedirectComponent(productNotFoundRedirectConfig);
+  }
+
+  // @TODO: add error handling
+  if (!product || error) {
+    return <div>Error...</div>;
```
dmtrth25, line 50: maybe for now good to hadnle some `<AppTypography translationKey="error.label" />` for error. 
like our previously examples.
Also nice to render some `AppLoader` for loading instead text above I mean to handle `isLoading`

OliaHrytsak (approved):

OliaHrytsak:
Excellent work! i only think about the way to awoid scrolling inside item container.
Maybe split into two columns?

heshvab (approved):

code/src/pages/product-details/ProductDetailsPage.tsx
mhevyk: Thanks, I moved it

code/src/pages/product-details/ProductDetailsPage.tsx
mhevyk: Thanks. It looks simpler in this case

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
mhevyk: Thanks for review! I don't think that is is good idea to move local types to separate file, but I managed to reuse `ProductDetailsPageParams` type from previous comment

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
mhevyk: Thank you for the question! I don't think that we need to create some util for it. I don't see any use case in another component for now. If I am wrong, please correct me

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.scss
mhevyk: Thanks for comment! I moved styles for image into previous style, now it looks better

mock-backend/src/routes/productRoutes.js
mhevyk: I guess we can create a low-priority issue for it. In my point of view, it is not so important for now because we use mock backend just for testing. Let's keep this comma, because the developer who will work on this issue will execute prettier format command and all files would be formatted at once using `.prettierrc`

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
D1nviz: I mean you can use in following:

type ProductDetailsContainerProps = {
  productId: ProductId; // wich is imported from `products.types`

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.tsx
mhevyk: As for error, @D1nviz assigned to an issue to create reusable error layout that will look nicer. I think we can also create issue for fixing all loader on all dashboard tabs, because for now on orders page for example we have only `<div>Loading...</div>` for now.

I attached a screenshot below:

image

Same for products:

image

code/src/pages/product-details/components/product-details-container/ProductDetailsContainer.test.tsx
mhevyk: Let's consider following test:

```tsx
test("renders description correctly", () => {
    renderAndMock({
      data: { ...mockProduct, description: "value1/value2/value3/value4" }
    });

    const descriptionParagraphs = screen.getAllByText(/value\d/);
    expect(descriptionParagraphs).toHaveLength(4);
  });
```

We receive a string from backend that includes slashes that separate different characteristics of product. I just wanted to test that if we have description "value1/value2/value3/value4" then we should have 4 rendered paragraphs. I tried to use it without regex, but i got the error related to keys:

image


That's why i decided to use regex to match different items. I think it is quite readable. If you have a better suggestion then I am open to them

Let's also consider following test:

```
const deliveryMethodImages = screen.getAllByAltText(
      /dashboardTabs.orders.filters.\w+/
    );
```

I used regex here because for all delivery method images we have prefix `dashboardTabs.orders.filters.` and it will reduce code if we want to add new delivery method. I also modified assertion like follows:

```
expect(deliveryMethodImages).toHaveLength(deliveryMethods.length);
```

With this approach we won't have to change this test if we add new delivery method and it will pass

mhevyk:
> Excellent work! i only think about the way to awoid scrolling inside item container. Maybe split into two columns?

Thanks! I appreciate it

I decided to use scroll here because without it product details page's section with image visually looks empty. For example on Rozetka we have a bunch of small images to preview:

image

In this case I was inspired by [comfy website](https://comfy.ua/ua/naushniki-vkladyshi-besprovodnye-tws-oppo-enco-air3-pro-green.html). Maybe you can make an advice on the design of this section, because I am not a perfect designer :)

D1nviz (approved):

mock-backend/src/routes/productRoutes.js
dmtrth25: > I guess we can create a low-priority issue for it. In my point of view, it is not so important for now because we use mock backend just for testing. Let's keep this comma, because the developer who will work on this issue will execute prettier format command and all files would be formatted at once using `.prettierrc`

Sure. ok

dmtrth25 (approved):

PR #422, opened 2024-07-30
Added update product route

### LINKED ISSUE #356: Create product updation route
Create product updation route and page (without any content)

heshvab:
- Added update product route and tests


code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    mockAndRender({
+      productId: validUUID,
+      rtkqResponse: { error: { status: 500 }, data: mockData }
+    });
+
+    const label = screen.getByText("errors.somethingWentWrong");
+    expect(label).toBeInTheDocument();
+  });
+
+  test("Should display loadign message while loading", () => {
```
mhevyk: Typo here (`loadign`)

code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <DashboardTabContainer>
+      <AppBox className="dashboard-products-tab__toolbar">
+        <AppTypography
+          component="h1"
+          variant="h3"
+          translationKey="product.update.title"
+        />
+      </AppBox>
+      {data?.id}
```
mhevyk: Why do we need it here?

code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+const DashboardUpdateProductPage = () => {
+  const { productId } = useParams();
+
+  const { renderRedirectComponent } = useErrorPageRedirect();
+
+  if (!productId || !isUuidValid(productId)) {
+    return renderRedirectComponent(updateProductPageNotFoundErrorConfig);
+  }
+
+  const { data, isLoading, error } = useGetManagerProductQuery({ productId });
```
mhevyk: Did you test this component? I am not sure but I think we can get error react rendered more hook comparing to previous render because we try to use hook after return

D1nviz (approved):

code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.tsx
heorhiiss: That is just to display something. In another PR I will add form there.

code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.tsx
heorhiiss: I believe currently it woudn't throw error because it depends on id in route and if we change id through browser search bar, entire page will be reloadeded. In other words, there are no possibility that the same component will be firstly rendered with that hook and after without it. But in future, there may be problems if we will add some additional routing logic to this page. I changed it, now there are two components

code/src/pages/dashboard/dashboard-update-product/DashboardUpdateProductPage.test.tsx
heorhiiss: Thanks, fixed

mhevyk (approved):

dmtrth25 (approved):

PR #428, opened 2024-07-31
Add content to Users tab on admin panel

### LINKED ISSUE #333: Add content to Users tab to admin panel
Access for this tab should have only admin

image

mhevyk:
I tried to implement this logic before actual endpoint implementation, so i guess we can finally sync this tab after completion this task on backend

image

image

image

image

image


code/src/constants/common.ts
```diff
@@ -17,3 +17,8 @@ export const ROLES = {
   SHOP_MANAGER: "ROLE_MANAGER",
   ADMIN: "ROLE_ADMIN"
 } as const;
+
+export const USER_STATUSES = {
+  ACTIVE: "ACTIVE",
+  DEACTIVATED: "DEACTIVATED"
+} as const;
```
dmtrth25, line 24: In general what do you think about `enum` approach ?
I thought about it but didn't catch why `const` obj with properties better ?
What do you think ?

code/src/containers/tables/users-table/components/users-table-body/UsersTableBody.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <AppTableCell>
+        <AppTypography variant="caption">{user.email}</AppTypography>
+      </AppTableCell>
+      <AppTableCell>{roleBadge}</AppTableCell>
+      <AppTableCell>
+        <AppTypography variant="caption">
+          {formatDate(user.createdAt)}
+        </AppTypography>
+      </AppTableCell>
+      <AppTableCell>{statusBadge}</AppTableCell>
```
dmtrth25: what do you think about destruction `user` properties ? `status` `role` above too.
And also render some variable instead `{user.firstName} {user.lastName}`

code/src/constants/common.ts
mhevyk: Discussed

code/src/containers/tables/users-table/components/users-table-body/UsersTableBody.tsx
mhevyk: Fixed

dmtrth25 (approved):

OliaHrytsak (approved):

D1nviz:
I suggest adding width property to these badges. Make them similar
image

code/src/containers/tables/users-table/components/users-table-body/UsersTableBody.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import AppBadge from "@/components/app-badge/AppBadge";
+import { AppTableCell } from "@/components/app-table/components";
+import AppTypography from "@/components/app-typography/AppTypography";
+
+import formatDate from "@/utils/format-date/formatDate";
+
+const UsersTableBody = ({ user }: UsersTableBodyProps) => {
+  const { firstName, lastName, role, status, createdAt, email } = user;
+  const fullName = `${firstName} ${lastName}`;
```
D1nviz: A full name is also used in another place. Maybe create function for contact first and last names? You can create the separate issue for that

ValKoDim (approved):

heshvab (approved):

mhevyk:
> I suggest adding width property to these badges. Make them similar image

Thanks a lot! I did as you suggested, now the width of badge is calculated based on longest badge translation

code/src/containers/tables/users-table/components/users-table-body/UsersTableBody.tsx
mhevyk: I created util and reused it

D1nviz (approved):

PR #431, opened 2024-07-31
Created debounce and optimistic logic for changing product quantity

### LINKED ISSUE #390: Add debouncing logic to cart page logic
https://github.com/user-attachments/assets/82ed578f-5042-4715-a2e8-2fb825b9a571

### LINKED ISSUE #391: Add optimistic updates for cart page
If we click fastly on plus or minus buttons, price is lagging and not updating correctly. Use optimistic update technique to immidiately see correct price

https://github.com/user-attachments/assets/4b3ffccd-511a-42bf-adbc-35f3753f4c5f

dmtrth25:
Implemented `debounce` for changing quantity and optimistic updates logic for `totalPrice` of cart

https://github.com/user-attachments/assets/4ecd2321-ef37-4145-957a-63b85af8a2a3

**Resolved: #390 #391**


OliaHrytsak (approved):

OliaHrytsak:
Looks great!

heshvab (approved):

mhevyk (commented):
Generally great work performed! I have tested it locally. With long response from server it works good but it fails when we get error from server. I modified code as follows:

image

In this case cart brokes in some reason:

https://github.com/user-attachments/assets/3576d718-fa2c-4567-90c8-3cdc33006936

Can you try to fix this case? For example our server can return an error for adding item to cart and our whole app breaks

It would be great to show maybe error snackbar and rollback value to initial. What do you think?

D1nviz (approved):

dmtrth25:
> Generally great work performed! I have tested it locally. With long response from server it works good but it fails when we get error from server. I modified code as follows:
> 
> image
> 
> In this case cart brokes in some reason:
> 
>  breaks.mp4 
> Can you try to fix this case? For example our server can return an error for adding item to cart and our whole app breaks
> 
> It would be great to show maybe error snackbar and rollback value to initial. What do you think?

I have some comments about that. We should to discuss
In my case with your backend changes looks like

https://github.com/user-attachments/assets/872a874c-bda6-4af6-bd49-af1e191c3a0d

dmtrth25:
@mhevyk when you tried to realize some server error it should be without `async await`

image

mhevyk (approved):

PR #435, opened 2024-08-02
Changes to skeleton footer container

### LINKED ISSUE #368: Fix Skeleton view for product item
When loading for the product items should to fix the Skeleton view.
For now it's look like:

image

ValKoDim:
Changed the skeleton footer, so it doesn't extend when viewport is widened.


code/src/components/product-skeleton/ProductSkeleton.tsx
```diff
@@ -16,7 +16,7 @@ const ProductSkeleton = () => {
         width={254}
         animation="pulse"
       />
-      <AppBox className="spa-product-skeleton__footer">
+      <AppBox className="spa-product-skeleton__footer" width={234}>
```
D1nviz: You should move this property to SCSS. Also, is it flexible with this hardcoded width on different screen sizes? Maybe max-width would be better?

OliaHrytsak:
Can you attach the wideo how it looks now?

code/src/components/product-skeleton/ProductSkeleton.tsx
ValKoDim: Yes it works as intended. The skeletons themselves have hardcoded values, so on different screen sizes, the container for the footer skeletons stretch out. having fixed width for the container fixes the issue. Moving this property to the SCSS, doesn't seem to work. It gets overwritten somehow.

mhevyk: Then try to use `!important`. We used it several times because styles of mui is a bit hard to override in terms of specificity

ValKoDim:
@OliaHrytsak @D1nviz 
This is how it looks now: 
https://github.com/user-attachments/assets/40a23277-3ff2-4b2a-a0cb-e3d5cfd6efda

code/src/components/product-skeleton/ProductSkeleton.tsx
ValKoDim: Done

dmtrth25 (approved):

D1nviz (approved):

code/src/components/product-skeleton/ProductSkeleton.scss
```diff
@@ -6,6 +6,7 @@
 
   &__footer {
     display: flex;
+    width: 234px !important;
```
mhevyk, line 9: Generally we add comment to inform that this is for overriding mui styles, because using `!important` is not a good practice. We discussed it with our team. Check how we did this in other places in the project (use global search in your IDE)

OliaHrytsak (approved):

heshvab (approved):

PR #438, opened 2024-08-02
Implemented product filtering by category using backend endpoint

### LINKED ISSUE #436: Replace current logic for filtering products by categories, with one using the backend.
Currently the filtering of the products by their categories is done in the frontend, we need to implement the endpoint created by the backend team for the filtering.

ValKoDim:
Refactored the logic behind products data recovery from the database to include the new endpoint for product filtering.  Also refactored some test to cover those changes. I have also refactored our mock backend to filter products by categories as well.


code/src/pages/products/ProductsPage.tsx
```diff
@@ -31,6 +30,7 @@ const ProductsPage = () => {
     isLoading,
     isError
   } = useGetUserProductsQuery({
+    tags: categoryType ? `category:${categoryType}` : "",
```
OliaHrytsak, line 33: Please attach the wideo to show how it works now?
With real backend

ValKoDim: Pretty much, the endpoint accepts a new property called tags. The tag is a string in the form of "category:<category>". If we pass this string we get all products that have this tag. If we pass an empty string, we get list of all products. It is a bit unintuitive, because we have to pass a string in this format and not just the name of the category, but that is how backend team have implemented it.

code/src/containers/best-sellers/BestSellers.test.tsx
```diff
@@ -55,7 +55,8 @@ describe("BestSellers", () => {
     expect(useGetUserProductsQuery).toHaveBeenCalledWith({
       page: 0,
       size: 5,
-      lang: "en"
+      lang: "en",
+      tags: ""
```
mhevyk, line 59: Why this is empty here? I suggest making `tags` optional property

code/src/pages/products/ProductsPage.test.tsx
```diff
@@ -46,6 +46,7 @@ jest.mock("@/context/i18n/I18nProvider", () => ({
 }));
 
 const defaultQueryArguments = {
+  tags: "",
```
mhevyk: Same

code/src/types/product.types.ts
```diff
@@ -35,6 +35,7 @@ export type GetUserProductsResponse = {
 };
 
 export type GetUserProductsParams = Lang & {
+  tags: string | null;
```
mhevyk: Do it here

mock-backend/src/controllers/productController.js
```diff
@@ -1,3 +1,5 @@
+const {categoryFilter} = require('../utils/categoryFilter')
```
mhevyk: This lacks formatting. Should be like `{ categoryFilter }`

code/src/containers/best-sellers/BestSellers.test.tsx
ValKoDim: Done

code/src/pages/products/ProductsPage.test.tsx
ValKoDim: Done

code/src/types/product.types.ts
ValKoDim: Done

mock-backend/src/controllers/productController.js
ValKoDim: Done

mhevyk (approved):

OliaHrytsak (approved):

D1nviz:
Did you try it with real backend?

mock-backend/src/controllers/productController.js
```diff
@@ -10,15 +12,18 @@ const getAllProducts = (req, res) => {
   const page = validateNumberQueryParam(req.query.page);
   const size = validateNumberQueryParam(req.query.size, 10);
   const { sort } = req.query;
+  const category = req.query.tags;
```
dmtrth25: you can get `tags` with `req.query` destructuring and if you want to rename it with `: category`

Look at

```suggestion
const { sort, tags: category } = req.query
```

mock-backend/src/utils/categoryFilter.js
```diff
@@ -0,0 +1,10 @@
+const categoryFilter = (category, productsList) => {
+  if (category === "") {
+    return productsList;
+  }
```
dmtrth25: use `if(!category)`

mock-backend/src/utils/categoryFilter.js
```diff
@@ -0,0 +1,10 @@
...
+  return productsList?.filter((product) => {
+    return product.tags.includes(category);
+  });
```
dmtrth25: it can be written in one row

mock-backend/src/utils/categoryFilter.js
```diff
@@ -0,0 +1,10 @@
...
+};
```
dmtrth25: Also in general it can write simplier like
```
const categoryFilter = (category, productsList) => !category ? productsList : productsList?.filter(product => product.tags.includes(category))

```

What do you think ?

ValKoDim: @mhevyk I had to revert some of the changes you suggested. The reason is that the backend expects to receive tags param always. if we don't  provide at least empty string, we get nothing. Empty string gets us all products, without category filtering.

ValKoDim:
@mhevyk I had to revert some of the changes you suggested. The reason is that the backend expects to receive tags param always. if we don't provide at least empty string, we get nothing. Empty string gets us all products, without category filtering.

ValKoDim:
> Did you try it with real backend?

@D1nviz Yes, it works on prod.

D1nviz (approved):

mock-backend/src/utils/categoryFilter.js
ValKoDim: Done

mock-backend/src/utils/categoryFilter.js
ValKoDim: I moved it all on 1 line.

mock-backend/src/utils/categoryFilter.js
ValKoDim: Done

mock-backend/src/controllers/productController.js
ValKoDim: I applied the suggested changes.

dmtrth25 (approved):

PR #461, opened 2024-08-06
Add pagination to users dashboard tab

### LINKED ISSUE #456: Add pagination for users dashboard tab

mhevyk:
Added pagination for users tab. I attached video how it works with prod for now

https://github.com/user-attachments/assets/bb0aac13-4e69-47ae-bd63-e4d2edbfbfe9

Also I refactored pagination logic on dashboard tabs so that it can be reused both on frontend and mock backend
Also I made a minor change to styles on orders tab (made font size of badge smaller)


heshvab (approved):

code/src/components/app-pagination/AppPagination.tsx
```diff
@@ -1,25 +1,30 @@
...
+const AppPagination = (props: AppPaginationProps) => {
+  const { page, setPage } = usePagination();
 
-    const itemSearchParamsString = searchParams.toString();
+  const renderPaginationItem = ({
+    onClick,
+    ...itemProps
+  }: Omit<PaginationRenderItemParams, "selected">) => {
+    const handlePageChange = (event: MouseEvent<HTMLDivElement>) => {
+      setPage(itemProps.page!);
```
D1nviz: Why did you use `!` here?

mhevyk: I did it because `itemProps.page` can be either `string | null`. If we receive `null`, it is not a problem because `validatePage` util takes care about this case. I adjusted this logic a little bit to make more understandable:

```tsx
if (newPage !== null) {
      params.set(PAGE_KEY, newPage.toString());
    } else {
      params.delete(PAGE_KEY);
    }
```

D1nviz (approved):

dmtrth25 (approved):

PR #465, opened 2024-08-06
Remove recommended sort param

### LINKED ISSUE #464: Remove recommended sort param
We need to remove recommended param because backend changed it so that recommended is default option on their side and in this case we should pass nothing

mhevyk:
- Removed default recommended param. When we select recommended sort, sort param is not present in request url it works correctly with backend. I tested it with real backend using `npm run start:prod` command
- Increased coverage with unit and mutation tests for `ProductsPage`

https://github.com/user-attachments/assets/3e5dcf43-49c4-47ab-8dbb-1a7575443124

image

image


OliaHrytsak:
Why are you decided to leave this Recomended parameter? Mayby to avoid this confusing and change this default label into ''Choose the option" for exaple and remove this parameter from sortOptions ?

mhevyk:
> Why are you decided to leave this Recomended parameter? Mayby to avoid this confusing and change this default label into ''Choose the option" for exaple and remove this parameter from sortOptions ?

I replicated an approach on popular retail shops. Let's consider Rozetka:

image

All retail shops have in common that that always have default sorting option and it is not called Choose and option, which in my point of view does not create any confusion

heshvab (approved):

D1nviz (approved):

OliaHrytsak (approved):

PR #475, opened 2024-08-07
Added searching for manager products page

### LINKED ISSUE #399: Add search field to search product by name
image

heshvab:
- Added search functionality for manager products
- Added tests for it

https://github.com/user-attachments/assets/5cc2cd8d-3a9a-479e-9881-59a371c010b7


code/src/pages/dashboard/dashboard-products/components/dashboard-products-header/DashboardProductsHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { formatMessage } = useIntl();
+
+  const handleSearchChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const value = e.target.value;
+    setSearchValue(value);
+  };
+
+  const handleClearSearch = () => setSearchValue("");
+
+  const handleSearch = () => {
```
OliaHrytsak, line 32: probably you will fased with the problem, that when you are not on the first page and ckick on handleSearch button, the table would be empty, because results are on the previous page
My deсision was change also page param into 1, to redirect user on first list page

code/src/pages/dashboard/dashboard-products/components/dashboard-products-header/DashboardProductsHeader.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  test("Should clear input value", async () => {
+    await typeIntoInput(input, value);
+
+    expect(input).toHaveValue(value);
+
+    const clearButton = screen.getByLabelText("clear");
+
+    fireEvent.click(clearButton);
+
+    await waitFor(() => {
```
OliaHrytsak: I think this async await is unnecessary there and on the previous test
In util typeIntoInput already is async/await logic
And for expect you can use only waitFor

heorhiiss: Thank you. I changed it according to your suggestion, it works without async in both tests

code/src/pages/dashboard/dashboard-products/components/dashboard-products-header/DashboardProductsHeader.tsx
heorhiiss: Thank you for notice. I haven't encountered such problem. I think that is because in your case hook for filtering is used. I will also check behaviour with real backend in few minutes, when it will be available.

dmtrth25 (approved):

mhevyk (commented):
What will happen if we clear search input? I can suppose that search param will persist but with empty value. I suggest deleting it from params in this case or use `useFiltersWithApply` logic for it

The same problem I guess has orders search field, so we can also fix it there

code/src/pages/dashboard/dashboard-products/components/dashboard-products-header/DashboardProductsHeader.tsx
heorhiiss: I checked it with real backend, everything works fine

OliaHrytsak (approved):

code/src/pages/dashboard/dashboard-products/DashboardProductsPage.tsx
```diff
@@ -20,13 +16,21 @@ const DashboardProductsPage = () => {
   const { locale } = useLocaleContext();
   const { page } = usePagination();
 
+  const [searchParams, setSearchParams] = useSearchParams({ search: "" });
+  const searchValue = searchParams.get("search") || undefined;
```
D1nviz, line 20: Do we need `|| undefined`?  `.get` returns null

heshvab:
> What will happen if we clear search input? I can suppose that search param will persist but with empty value. I suggest deleting it from params in this case or use `useFiltersWithApply` logic for it
> 
> The same problem I guess has orders search field, so we can also fix it there

fixed, take a look

mhevyk (approved):

code/src/pages/dashboard/dashboard-products/DashboardProductsPage.tsx
heorhiiss: I think yes because empty value like `searchByName=&page=1` was settled before I added `undefined`

D1nviz (approved):

PR #485, opened 2024-08-08
Added e2e tests for product form

### LINKED ISSUE #434: Add e2e tests to manager product forms

heshvab:
- Added e2e tests for new product form and create product form
- Fixed downed coverage in some files I previously change, now the coverage is back `100%`
- Fixed some imports
- Fixed bug with cart synchronization. Now cart is synchronized only for simple user, because previously when you save In cart items and authorize as manager, error was shown.


OliaHrytsak (approved):

code/src/pages/dashboard/dashboard-orders/DashboardOrdersPage.test.tsx
```diff
@@ -182,6 +182,23 @@ describe("DashboardOrdersPage", () => {
...
 
+  test("resets search param if input is submited without value", async () => {
+    renderAndMock();
+
+    const searchInput = screen.getByPlaceholderText(
+      /dashboardTabs.orders.search/
+    );
+
+    await typeIntoInput(searchInput, "user");
+    await typeIntoInput(searchInput, "");
```
mhevyk, line 193: Why do we need to type one value and then another?

code/src/pages/dashboard/dashboard-products/DashboardProductsPage.test.tsx
```diff
@@ -137,4 +137,23 @@ describe("DashboardProductsPage", () => {
       });
     });
   });
+
+  test("Should call useGetManagerProductsQuery with searchByName equal to undefined when input is empty", async () => {
```
mhevyk: `async` can be removed i guess

code/src/pages/dashboard/dashboard-orders/DashboardOrdersPage.test.tsx
heorhiiss: This test checks if the function is called, that function is called when user clicks search button with empty input. For some reason, it doesn't work other way. I tried to omit typeIntoInput, because input should be empty by default, and also tried yo use only `await typeIntoInput(searchInput, "");`, without previously setting `user.` Both ways doesn't work for some reason

code/src/pages/dashboard/dashboard-products/DashboardProductsPage.test.tsx
heorhiiss: Thanks for noticing, fixed

D1nviz (approved):

mhevyk (approved):

